PR #421, opened 2026-03-16
Integrate Remotion with MCP video editor

tombeckenham:
https://claude.ai/code/session_017gou586U3vg67ovYks2jDT


tombeckenham:
Not gong to do this as remotion doesn't have the sort of license we need. I'm probably going to use mediabunny

PR #593, opened 2026-04-20
Upload reference elements for sequences

### LINKED ISSUE #271: Upload Elements
Allow the user to upload any number of images when creating a new sequence. They should be able to refer to these by file name when editing the script by including the filename in caps lock. Enhancing the script should be able to use vision apis to understand what has been uploaded and refer to them automatically during the enhance process. Then these elements should behave like characters or locations somewhat. An element bible should be created against the sequence when scene splitting. This should indicate when an element is used. A link to the element image should be included when generating images or videos if the element is referenced in a scene or prompt.

Elements would usually be something like a logo,a screenshot, or a product image.

tombeckenham:
## Summary
- Per-sequence reference image uploads (logos, product shots, screenshots) referenced by UPPERCASE tokens in scripts
- New Elements tab in sequence detail + new Elements tab per scene
- Drag-drop anywhere in the script editor (handles File drags, browser image drags via `text/uri-list`/`text/html`, and Finder preview)
- Script enhancement now passes reference images to vision models via @tanstack/ai content parts; raw-fetch bypass removed and Langfuse observation properly linked (prompt version, teamId, userId, tags)

## Test plan
- [ ] Create a sequence with uploaded elements; verify tokens appear in the enhanced script and in each scene's Elements tab
- [ ] Drag an image from macOS Finder, the Finder preview pane, and another browser tab — all should upload
- [ ] Regenerate from an existing sequence's script tab with elements visible
- [ ] Confirm Langfuse shows a `script-enhance` observation linked to the prompt, with teamId/userId


src/functions/sequence-elements.ts
```diff
@@ -0,0 +1,207 @@
...
+  while (
+    (await scopedDb.sequenceElements.getByToken(sequenceId, candidate)) !== null
+  ) {
+    candidate = `${token}_${suffix}`;
+    suffix += 1;
+    if (suffix > 100)
+      throw new Error('Unable to generate unique element token');
+  }
+  return candidate;
+}
```
tombeckenham: I'm thinking this could be added to scopedDb itself

src/components/element/element-selector.tsx
```diff
@@ -0,0 +1,495 @@
...
+        .filter((u): u is DraftElementUpload => !!u),
+    [entries]
+  );
+
+  // Report drafts to parent (draft mode only)
+  useEffect(() => {
+    if (isPersisted) return;
+    props.onDraftElementsChange?.(uploaded);
+  }, [isPersisted, uploaded, props.onDraftElementsChange]);
+
```
tombeckenham: I don't love this pattern. Is there a way to do this without using useEffect / having a funciton as a dep?

PR #646, opened 2026-05-01
Stage 1: migrate remaining content-generation workflows to snapshots #617

### LINKED ISSUE #617: Stage 1: migrate remaining content-generation workflows to snapshots
Rolls the snapshot pattern from #615 across the rest of the content-generation workflows, per [workflow-snapshots-and-content-hash-staleness](docs/architecture/workflow-snapshots-and-content-hash-staleness.md) § "Per-workflow input surface".

Depends on #615 and #616.

## Workflows to migrate

For each: inline the missing hashes at trigger time, remove any live mutable reads inside `context.run`, opt into `createScopedWorkflow`'s `snapshot` config, and route divergent results per the rules below.

- **`characterSheetWorkflow`** — add talent-sheet `input_hash` to the payload. Divergence: discard and re-queue (Stage 1 has no character variants table — that's Stage 2, #617).
- **`locationSheetWorkflow`** — add library-location `reference_input_hash` to the payload. Divergence: discard and re-queue.
- **`libraryTalentSheetWorkflow`** — payload already inlines `referenceImageUrls`/`talentDescription`; add a `snapshotInputHash`. Divergence: discard and re-queue.
- **`frameImagesWorkflow`** — alongside each character/location URL in the payload, inline its sheet `input_hash`. Divergence: write to `frame_variants` (same as #615).

Each migrated workflow should also write its computed `input_hash` into the matching column on the entity it produces (`characters.sheet_input_hash`, `location_sheets.input_hash`, etc.) — this is what makes downstream staleness detection work.

## Acceptance

- All four workflows opt into `context.snapshot`.
- Convergent writes populate the new `*_input_hash` columns.
- Sheet workflows that detect divergence emit `stale:detected` and re-queue with current inputs.
- `frameImagesWorkflow` divergence lands in `frame_variants` like #615.

tombeckenham:
## Related Issue
Closes #617

> Stacked on `616-stage-1-realtime-stale-detected-event-ui-surface`. Merge the parent PR first; GitHub will auto-retarget this one.

## Summary
Migrated characterSheet/locationSheet/libraryTalent/frameImages workflows to snapshot pattern with input-hash writes + divergence handling; added sequence_locations.reference_input_hash column

## Notes


tombeckenham (commented):
## Review: Stage 1 snapshot migration — content-generation workflows

Solid migration overall — the snapshot config wiring, hash payload design (per-scene snapshots in `frameImagesWorkflow`, talent-hash inlining in `characterSheetWorkflow`), and divergence-aware reconcile pattern all line up with the established pattern in `scoped-workflow.ts`. Three blocking concerns and a handful of important follow-ups.

### Critical

**1. Divergent path returns a phantom URL the DB never adopts**

`src/lib/workflows/character-sheet-workflow.ts:204-211` and `src/lib/workflows/location-sheet-workflow.ts:200-205`: when `reconcileOutcome === 'divergent'`, the workflow correctly skips the DB update and re-queues, but still returns `{ sheetImageUrl: storageResult.url, sheetImagePath: storageResult.path, ... }` to the caller.

`src/lib/workflows/recast-character-workflow.ts:84-90` and `src/lib/workflows/recast-location-workflow.ts:82-86` then guard on `!sheetResult?.sheetImageUrl` — which passes — and proceed to drive `regenerateFramesWorkflow` with that URL. Net effect:

- Caller (UI/realtime) sees recast "succeed" with URL `X`.
- DB still holds the previous sheet (or none).
- Frame regeneration runs against the stale DB-resident sheet.
- The re-queued sheet workflow later writes a *different* image `Y` to the DB.

Suggested fix: throw a typed `SnapshotDivergedError` from the divergent branch (or re-queue the parent recast as well) so the recast pipeline aborts instead of advertising an orphaned URL.

**2. Re-queue can infinite-loop with billing amplification**

Same three sheet workflows. `deduplicationId` includes `${snapshotInputHash.slice(0, 16)}` — but the hash changes each iteration when upstream keeps mutating, so QStash accepts every re-queue. Each iteration runs `generateImageWithProvider` and `deductWorkflowCredits` *before* the divergence check, so a thrashing upstream burns credits indefinitely. There is no `requeueDepth` counter on the input and no log warning. Add a bounded depth or an exponential backoff.

**3. `sheet-snapshots.ts` — 235 lines of hash logic with zero tests**

Hash determinism is the load-bearing contract for the entire divergence-detection system. A determinism bug here makes divergence either fire on every run (silent over-regeneration, billing amplification) or never fire (silent stale writes). Highest-priority cases to pin:

- **FromDto/Current parity** for each of the four helper pairs (build payload → hash via `FromDto` → mock matching DB → hash via `Current` → must be equal).
- **Sort asymmetry**: `computeLibraryTalentSheetHashFromDto` (line ~167) hashes `input.referenceImageUrls` *unsorted*, while `computeLibraryTalentSheetHashCurrent` (line ~184) sorts the live media URLs. Callers in `src/functions/talent.ts:140,413` pre-sort, but a future caller that forgets will see false divergence on every run. Sort inside `FromDto` too.
- `computeStyleConfigHash(null) === computeStyleConfigHash(undefined) === 'no-style'`.
- `imageModel ?? DEFAULT_IMAGE_MODEL` agreement across both paths (a bump to `DEFAULT_IMAGE_MODEL` would otherwise diverge every payload that omits the field).

### Important

**4. `?? 'unknown'` sentinel leaks into the realtime channel**

`character-sheet-workflow.ts:217` and `location-sheet-workflow.ts:217`: `snapshotInputHash: context.snapshot?.snapshotInputHash ?? 'unknown'`. The divergent branch is only reachable when `context.snapshot` is set (the comparison above guarantees it), so the sentinel is dead. But it teaches downstream consumers that `'unknown'` is a legal hash value. Narrow above the emit and drop the fallback.

**5. Library-talent re-queue silently uses stale URLs when talent vanishes**

`library-talent-sheet-workflow.ts:199-206`: `talent?.media…sort() ?? input.referenceImageUrls ?? []`. If the talent row was deleted, this re-queues with the stale payload (or `[]`), guaranteeing another divergence and another billable image generation. Hard-fail when `talent == null` in the re-queue path (the documented fallback in `computeLibraryTalentSheetHashCurrent` is fine — that one is read-only).

**6. `frame-images-workflow.ts` `computeCurrent` is `computeFromDto` aliased**

`frame-images-workflow.ts:233-237`: both helpers hash the same inlined `sceneSnapshots`, so `validate()` only catches payload tampering, never genuine upstream drift. The comment at 220-227 acknowledges this; consider renaming or annotating the call site so future maintainers don't assume drift detection is happening here.

**7. Divergence path drops the diagnostic data needed to debug it**

`character-sheet-workflow.ts:182-205`, `location-sheet-workflow.ts:182-205`: when `currentHash !== snapshotInputHash`, the function returns `'divergent'` without logging either hash. A `console.warn('[CharacterSheetWorkflow] divergence detected', { snapshot, current, characterDbId })` before the return would make incident triage tractable.

**8. Orphaned R2 objects on divergence**

The freshly-uploaded image in the divergent branch is not referenced by the DB and not deleted. At minimum log the orphan path with a correlation ID; ideally schedule a cleanup.

**9. Duplicated upstream-hash resolution**

The talent-sheet hash lookup is open-coded in three places (`recast-character-workflow.ts:39-72`, `character-sheet-workflow.ts:225-236`, `sheet-snapshots.ts:computeCharacterSheetHashCurrent`) and the library-location-hash lookup similarly. Extract `resolveTalentSheetHash(scopedDb, characterId)` / `resolveLibraryLocationReferenceHash(...)` so the three callers can't drift.

### Nits

- `src/lib/workflow/types.ts:185-188`: doc-comment says "Snapshot pattern only — see docs/architecture/workflow-snapshots-and-content-hash-staleness.md" which is helpful; the other new fields could use the same pointer.
- `analyze-script-workflow.ts:312-329`: `frameImagesPayload.snapshotInputHash` is mutated after construction; consider building it once via a helper to keep the DTO immutable in spirit.

### Verdict

Three blockers (divergent return value, unbounded re-queue, missing hash tests) — please address before merge.

tombeckenham (commented):
## Round 2 review — snapshot migration

Round 1 blockers were resolved cleanly: divergent path now throws `SnapshotDivergedError` instead of returning the phantom URL, `MAX_REQUEUE_DEPTH=5` caps the re-queue loop with depth tracking on the input, sort is now applied inside `computeLibraryTalentSheetHashFromDto`, the `?? 'unknown'` sentinel is gone, divergence logs both hashes plus the orphaned R2 path, and `sheet-snapshots.test.ts` is in. Two new round-2 issues turned up while tracing the divergence semantics through to callers.

### Critical

**Recast parents convert child divergence into a hard `recast:failed` event** — `src/lib/workflows/recast-character-workflow.ts:79-83` and `src/lib/workflows/recast-location-workflow.ts:79-83`. When the child sheet workflow throws `SnapshotDivergedError`, `context.invoke` returns `isFailed: true`. The sheet's own `failureFunction` correctly skips the `:failed` realtime emit (`character-sheet-workflow.ts:320-326`), but the recast parent then throws a plain `new Error('Character sheet generation failed for ...')`, which has no divergence marker and propagates into the recast `failureFunction` at `recast-character-workflow.ts:197-216` / `recast-location-workflow.ts:196-` which emits `generation.recast:failed` to the UI. Net effect: user sees a hard recast failure, **and** the affected frames are never regenerated (recast aborted before `regenerate-frames`) even after the re-queued sheet succeeds. Either `isSnapshotDivergedFailure(failResponse)` should be checked in the recast parent (no-op or self-re-queue), or the divergence marker should be propagated through the thrown error so recast's `failureFunction` can suppress the `:failed` emit.

### Important

**`MAX_REQUEUE_DEPTH` exhaustion silently abandons the entity** — `character-sheet-workflow.ts:236-244`, `location-sheet-workflow.ts` (analogous), `library-talent-sheet-workflow.ts:317-323`. When the cap is hit, the inner `context.run` returns without re-queuing, but the outer `throw new SnapshotDivergedError(...)` still fires. The sheet `failureFunction` then treats the marker as transient and skips the `:failed` status update — so the entity is left in `generating` state forever, with no DB status change, no realtime failure event, and no operator signal beyond a `console.warn`. Either throw a distinct (non-marker) hard error when the cap is reached so the regular failure path runs, or explicitly call `updateSheetStatus(..., 'failed', 'diverged-loop')` before throwing.

**Hash parity tests are tautological for the convergent case** — `sheet-snapshots.test.ts:86-105, 165-180, 211-229`. `computeCharacterSheetHashCurrent` (and the location/talent equivalents) internally delegates to `computeXxxFromDto` after resolving upstream (`sheet-snapshots.ts:138-143, 178-185, 222-228`). So "FromDto and Current produce identical hashes when the resolved upstream matches the input upstream" passes by construction — a refactor that broke DTO field ordering would break both paths identically and the parity test would still pass. The divergence-direction tests do exercise the resolver branch and remain valuable. Suggestion: pin one fixed DTO to a fixed expected hash hex (snapshot test) so a wire-format change is detected.

### Nits

- `sheet-snapshots.test.ts:50-53` — `as unknown as ScopedDb` cast: fine for test stubs, but the inline comment "future helper that reaches deeper fails loudly" is incorrect — unstubbed property access returns `undefined` and the test crashes inside the helper, not at the boundary.
- `frame-images-workflow.ts:233-238` — `computeCurrent` is now correctly annotated as a payload-integrity-only alias of `computeFromDto`; `validate()` runs the hash twice as a result. Consider whether the snapshot config can express "no upstream resolution" without the duplicate call, but not worth blocking.

### Verdict

Round 2 cleared the original three blockers. The recast-divergence propagation gap is the one new blocker — it converts a self-healing transient signal into a user-visible recast failure with un-regenerated frames.

PR #648, opened 2026-05-02
Stage 1.5: per-frame snapshot validation + frame_variants divergence in generateImageWorkflow #647

### LINKED ISSUE #647: Stage 1.5: per-frame snapshot validation + frame_variants divergence in generateImageWorkflow
Follow-up to #617. Closes the gap in #617's last acceptance bullet (\"`frameImagesWorkflow` divergence lands in `frame_variants` like #615\") that was deferred to keep #617 reviewable.

## Problem

`frameImagesWorkflow` (#617) opts into the snapshot pattern only at the **batch boundary** — `computeCurrent` is aliased to `computeFromDto`, so `context.snapshot.validate()` catches payload tampering but not genuine upstream drift. The PR comment explicitly defers real drift detection to the downstream `generateImageWorkflow`:

> // NOTE: \`computeCurrent\` is intentionally an alias of \`computeFromDto\`
> // here ... Real drift detection happens per-frame downstream in
> // \`generateImageWorkflow\`, which re-resolves sheet hashes at write
> // time and routes divergent frames into \`frame_variants\`.

But `generateImageWorkflow` has **no snapshot wiring today** — \`grep snapshotInputHash src/lib/workflows/image-workflow.ts\` returns nothing. As shipped, a sheet regenerated mid-flight while `frameImagesWorkflow` is running silently overwrites the primary frame thumbnail rather than landing in `frame_variants` like #615 did for `regenerateFramesWorkflow`.

## Scope

1. **Pass per-frame snapshot context through the invoke**
   - `frameImagesWorkflow` already has `sceneSnapshots` on its payload with character/location/element sheet hashes per scene. Extend `ImageWorkflowInput` to carry the matching `FrameImageSceneSnapshot` for the single scene this image-workflow run owns.
   - `frame-images-workflow.ts` passes the per-scene snapshot when it does `context.invoke(generateImageWorkflow, …)`.

2. **Wire `generateImageWorkflow` snapshot config** (`src/lib/workflows/image-workflow.ts`)
   - `computeFromDto`: hash the inlined per-scene snapshot via `computeFrameImageSceneHash` (already exists in `sheet-snapshots.ts`).
   - `computeCurrent`: re-resolve character-sheet/location-sheet `input_hash` columns from live DB at write-time and recompute the same hash.
   - On divergence: write to `frame_variants` with `diverged_at` set + `input_hash` of the snapshot the alternate was scoped to (use the existing `frameVariants.insertDivergent` helper from #615). Do NOT touch the primary frame artifact.
   - On convergence: existing primary-slot dual-write, plus persist `input_hash` on the `frame_variants` primary row.

3. **Tests**
   - Convergent path: primary `frame_variants` row gets `input_hash` populated.
   - Divergent path (mock a sheet hash drift between trigger and write): a new `frame_variants` row is inserted with `diverged_at`, primary row untouched.

## Acceptance

- `generateImageWorkflow` opts into `context.snapshot` with non-aliased `computeFromDto` / `computeCurrent`.
- Convergent writes populate `input_hash` on the primary `frame_variants` row.
- Divergent writes land in `frame_variants` with `diverged_at`, mirroring #615's `regenerateFramesWorkflow` behavior.
- `frameImagesWorkflow`'s batch-level snapshot can have its `computeCurrent` properly implemented (re-resolving live sheet hashes per scene) once the per-frame path is the source of truth.

## References

- Architecture doc: `docs/architecture/workflow-snapshots-and-content-hash-staleness.md` § "Pillar 3: Divergence-on-completion"
- #615 — pattern reference (`regenerateFramesWorkflow` writes divergent into `frame_variants`)
- #617 — parent (deferred this bullet)
- #625 — UI consumer (will surface the alternates this issue produces)

tombeckenham:
## Related Issue
Closes #647

## Summary
Implemented per-frame snapshot validation and frame_variants divergence routing in generateImageWorkflow; pushed branch for PR + CI/review.

## Notes

### DIFF
--- a/src/lib/workflow/types.ts
+++ b/src/lib/workflow/types.ts
@@ -57,6 +57,21 @@ export interface ImageWorkflowInput extends SequenceWorkflowContext {
   referenceImages?: ReferenceImageDescription[];
   /** Skip R2 upload and store fal.ai CDN URL directly (for ephemeral preview images) */
   skipStorage?: boolean;
+  /**
+   * Per-scene snapshot for divergence detection. When present, the workflow
+   * re-resolves character/location/element sheet hashes at write time and
+   * routes divergent results into `frame_variants` instead of overwriting
+   * the primary thumbnail. Optional: omit for callers that handle their own
+   * divergence (e.g. `regenerateFramesWorkflow`) or for preview-mode runs.
+   */
+  sceneSnapshot?: FrameImageSceneSnapshot;
+  /**
+   * Aspect ratio frozen at trigger time. Required when `sceneSnapshot` is
+   * present so write-time hash recomputation matches the trigger-time hash.
+   */
+  aspectRatio?: AspectRatio;
+  /** Hash over `(prompt, model, aspectRatio, sceneSnapshot)`; validated at start. */
+  snapshotInputHash?: string;
 }
 
 /**
--- a/src/lib/workflows/divergence-writes.ts
+++ b/src/lib/workflows/divergence-writes.ts
@@ -1,0 +1,34 @@
+import type { NewFrame, NewFrameVariant } from '@/lib/db/schema';
+
+/**
+ * Reverts to apply when a snapshot-pattern image workflow detects divergence
+ * after speculatively writing the primary `frames` + primary `frame_variants`
+ * row at start. Each consumer pairs this with its own divergent-alternate
+ * INSERT payload.
+ */
+export function buildDivergentRevertWrites(): {
+  frame: Partial<NewFrame>;
+  primaryRevert: Partial<NewFrameVariant>;
+} {
+  return {
+    frame: {
+      thumbnailUrl: null,
+      thumbnailPath: null,
+      thumbnailStatus: 'pending',
+      thumbnailWorkflowRunId: null,
+      thumbnailGeneratedAt: null,
+      thumbnailError: null,
+      thumbnailInputHash: null,
+    },
+    primaryRevert: {
+      url: null,
+      storagePath: null,
+      previewUrl: null,
+      status: 'pending',
+      workflowRunId: null,
+      generatedAt: null,
+      error: null,
+      inputHash: null,
+    },
+  };
+}
--- a/src/lib/workflows/frame-images-workflow.ts
+++ b/src/lib/workflows/frame-images-workflow.ts
@@ -28,6 +28,7 @@ import {
   matchLocationsToScene,
 } from './scene-matching';
 import {
+  computeFrameImageSceneHash,
   computeFrameImagesHashFromDto,
   type FrameImageSceneSnapshot,
 } from './sheet-snapshots';
@@ -139,9 +140,27 @@ export const frameImagesWorkflow = createScopedWorkflow<
           ...elementRefs,
         ];
 
+        // Locate the per-scene snapshot inlined into our payload. Bound to
+        // sceneId rather than index so a re-ordered `sceneSnapshots` array
+        // (e.g. analyze-script sorts by sceneId; we don't) still maps right.
+        const sceneSnapshot = input.sceneSnapshots?.find(
+          (s) => s.sceneId === scene.sceneId
+        );
+
         // Generate with each selected model in parallel
         const modelResults = await Promise.all(
           imageModels.map(async (model) => {
+            // Compute the per-frame hash here (rather than at trigger time)
+            // because each model gets its own image-workflow invocation, and
+            // `computeFrameImageSceneHash` is keyed on the model.
+            const perFrameSnapshotInputHash = sceneSnapshot
+              ? await computeFrameImageSceneHash(
+                  sceneSnapshot,
+                  model,
+                  aspectRatio
+                )
+              : undefined;
+
             const result = await context.invoke(
               `image-${scene.sceneId}-${model}`,
               {
@@ -153,11 +172,14 @@ export const frameImagesWorkflow = createScopedWorkflow<
                   prompt: visualPrompt,
                   model,
                   imageSize,
+                  aspectRatio,
                   numImages: 1,
                   frameId: matchedFrame?.frameId,
                   sequenceId,
                   referenceImages:
                     allReferences.length > 0 ? allReferences : undefined,
+                  sceneSnapshot,
+                  snapshotInputHash: perFrameSnapshotInputHash,
                 } satisfies ImageWorkflowInput,
                 retries: 3,
                 retryDelay: 'pow(2, retried) * 1000',
--- a/src/lib/workflows/frame-images-workflow.ts
+++ b/src/lib/workflows/frame-images-workflow.ts
@@ -108,6 +108,12 @@ export const frameImagesWorkflow = createScopedWorkflow<
 
     const imageSize = aspectRatioToImageSize(aspectRatio);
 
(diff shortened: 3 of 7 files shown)


tombeckenham (commented):
## Automated review — Stage 1.5 snapshot validation

Reviewed `src/lib/workflow/types.ts`, `src/lib/workflows/{image-workflow,image-workflow-snapshot,image-workflow-snapshot.test,frame-images-workflow}.ts`. The hashing helpers are well-factored and tests pin the right contract dimensions (model, character/location/element drift, deterministic hashing). A few important issues to consider before merge — none block correctness on the happy path but several can hide real bugs.

### Important

**1. `aspectRatio` silently defaults inside the snapshot helpers**
`src/lib/workflows/image-workflow-snapshot.ts:49,68` falls back to `DEFAULT_ASPECT_RATIO` when missing, but `src/lib/workflow/types.ts:71-74` declares it required when `sceneSnapshot` is present. If a future caller forgets `aspectRatio` while passing `sceneSnapshot`, trigger-time and validate-time hashes silently diverge and the workflow throws with a confusing tamper-detection error. Per CLAUDE.md ("don't create fallback code — display errors to user"), throw `WorkflowValidationError` when `sceneSnapshot` is present but `aspectRatio` is missing.

**2. Divergent revert does not null `thumbnailUrl` / `thumbnailPath`**
`src/lib/workflows/image-workflow.ts:189-199` reverts lifecycle fields to `pending` but leaves `thumbnailUrl`/`thumbnailPath` untouched. The comment claims the row never held a URL — true on first generation, but if this workflow is ever reused on an already-completed frame, status flips to `pending` while the stale URL persists. Either explicitly null both fields in the revert or assert the precondition.

**3. `persist-result` step is non-idempotent on QStash retry**
`src/lib/workflows/image-workflow.ts:158-315` bundles storage upload + divergence check + frame update + variant write + realtime emit in a single `context.run`. A transient emit/DB failure now retries the storage upload, leaking R2 objects (the path may include a UUID/timestamp). `insertDivergent` is idempotent on `(frame, type, model, inputHash)` ✓, but `uploadImageToStorage` is not memoized. Consider splitting upload into its own step so the URL is durable, then a second step for divergence-routed DB writes.

**4. `previewUrl` reset is asymmetric**
`src/lib/workflows/image-workflow.ts:218` clears `previewUrl` on the divergent revert but the convergent path (`:286-299`) does not. If a prior preview-mode run populated `frame_variants.previewUrl`, the convergent completed write should also clear it.

**5. "Convergent on missing frame metadata" hides real corruption**
`src/lib/workflows/image-workflow-snapshot.ts:64-67`: `if (!frame?.metadata) return computeFrameImageSceneHash(...DTO...)`. This conflates a deleted frame (intentional) with a row that has `metadata = null` (data corruption). Convergent collapse then writes thumbnail to a frame whose scene metadata is gone. Distinguish `frame === undefined` from `frame.metadata == null` and log/throw on the latter.

**6. Test coverage gap on the critical path**
The new tests pin the two pure hashers comprehensively, but the divergent/convergent persist branches in `image-workflow.ts:158-315` and the snapshot+hash propagation in `frame-images-workflow.ts:143-182` are not exercised at all. The divergent revert+`insertDivergent`+emit sequence is exactly the bug class this PR exists to prevent — a workflow-level test mocking `context.run`/`context.snapshot`/`scopedDb` to assert (a) divergent → `frame_variants.insertDivergent` called, primary `frames` reverted, and (b) convergent → primary `frames` updated, no `insertDivergent` call, would lock that contract.

### Nits

- `src/lib/workflows/image-workflow.ts:267,297` — `inputHash: snapshotOpted ? (snapshotHash ?? null) : null`. By the time we're in convergent write, `snapshotOpted && snapshotHash` is guaranteed by the validator; the `?? null` masks an invariant. Use `snapshotOpted ? snapshotHash : null`.
- `src/lib/workflows/image-workflow.ts:240,296` — `simpleHash(input.prompt)` is computed twice across the two branches; hoist once at step entry.
- `src/lib/workflows/frame-images-workflow.ts:146-148` — `find` per scene per model is O(scenes × models × snapshots). Build a `Map<sceneId, snapshot>` once before the outer loop.
- `src/lib/workflows/image-workflow.ts` divergent log — include `currentHash.slice(0,8)` alongside the snapshot hash so future debugging can identify which dimension diverged.

tombeckenham:
## Review items addressed

Follow-up to the [self-review above](https://github.com/openstory-so/openstory/pull/648#pullrequestreview-203079011). All 6 important items + 4 nits are resolved across two commits.

### Important — `aab8dc36`
1. **`aspectRatio` silent default** → new `requireAspectRatio()` throws `WorkflowValidationError` when `sceneSnapshot` is set but `aspectRatio` is missing (`image-workflow-snapshot.ts`).
2. **Stale `thumbnailUrl`/`thumbnailPath` on divergent revert** → both fields now nulled in the divergent frame writes (`buildImageDivergentWrites`).
3. **`persist-result` non-idempotent on QStash retry** → split into `upload-image` step (durable url+path) → `persist-result` step (divergence + DB writes + emit). R2 leak risk on retry eliminated.
4. **Asymmetric `previewUrl` reset** → convergent variant write also clears `previewUrl` now (`buildImageConvergentWrites`).
5. **Missing-frame vs null-metadata conflation** → `computeImageWorkflowHashCurrent` now distinguishes `frame === undefined` (deletion race → fall back to DTO hash) from `frame.metadata == null` (data corruption → throw `WorkflowValidationError`).
6. **Workflow-level test coverage gap** → extracted `persistImageResult()` orchestration so the call sequence is testable. New tests in `image-workflow-snapshot.test.ts` assert: divergent → primary frames reverted + insertDivergent called + emit `pending`; convergent → primary updated + emit `completed` + **no** insertDivergent; non-snapshot mode; frame-deleted short-circuit.

### Nits — `aab8dc36`
- Dropped `?? null` from `inputHash` assignments by narrowing `snapshotHash: string | null` once after `validate-snapshot`.
- Hoisted `simpleHash(input.prompt)` once at persist-step entry.
- Pre-built `Map<sceneId, snapshot>` once in `frame-images-workflow.ts` (was O(scenes × models × snapshots) `find`).
- Divergent log now includes `currentHash.slice(0,8)` alongside the snapshot hash.

### Simplify pass — `a1b6dc60`
- Extracted shared `buildDivergentRevertWrites()` into new `divergence-writes.ts` — both `regenerate-frames-snapshot.ts` and `image-workflow-snapshot.ts` now compose the revert from one place (kills ~14 duplicated lines).
- Inlined the redundant `divergent` boolean; enriched `PersistImageOutcome.divergent` to carry `snapshotHash` so the workflow body doesn't re-check.
- Trimmed narrating / step-numbering comments.
- Replaced `as unknown as ScopedDb` test cast by narrowing the production function input types (`ImageHashScopedDb`, `PersistImageScopedDb`, exported `SceneForHash` for the metadata read-shape). Production `ScopedDb` is a structural superset; tests build literals that satisfy the narrow types directly.
- Replaced the spy's `args: unknown[]` (which forced cast-to-read) with typed per-method call lists.

### Verification
- `bun typecheck` clean
- `bun test` — 529 pass / 0 fail / 3 pre-existing skips across 40 files
- Pre-commit hooks (format, lint, typecheck) clean on both commits
- Zero `as` casts in the changed files

PR #650, opened 2026-05-03
Stage 4: prompt versioning and variants #620

### LINKED ISSUE #620: Stage 4: prompt versioning and variants
Adds input-hash and variants treatment to prompts (visual, motion, music) so they have history and staleness independent of the rendered artifact. See [workflow-snapshots-and-content-hash-staleness](docs/architecture/workflow-snapshots-and-content-hash-staleness.md) § "Stage 4: prompt versioning".

Depends on Stage 1 (#617). Independent of Stages 2 and 3.

## Why

Prompts are themselves generated artifacts (from scene metadata, character/location bibles, style config, analysis model). Today they have **no history** — user edits to `frames.imagePrompt` / `motionPrompt` and `sequences.musicPrompt`/`musicTags` overwrite the AI version with no way back.

## Scope

1. New tables:
   - `frame_prompt_variants` — one row per prompt revision. Columns: `frameId`, `promptType` (`'visual' | 'motion'`), `text`, `components` (json), `parameters` (json), `source` (`'ai-generated' | 'user-edit' | 'regenerated'`), `input_hash` (over the upstream context that produced an AI prompt; null for user edits), `analysisModel`, `created_at`, `created_by`.
   - `sequence_music_prompt_variants` — same shape, parent `sequenceId`, `promptType` `'music'`, fields capture both `prompt` and `tags`.
2. Read path stays the same: `frame.imagePrompt`/`motionPrompt` and `sequences.musicPrompt`/`musicTags` remain as cached "current" pointers — no caller change.
3. Write path goes through a helper that inserts a variant row and updates the cached column atomically.
4. Add `*_prompt_input_hash` columns on `frames` for visual and motion prompts. Hash is computed from scene metadata + style config + character/location bible + analysis model. When upstream context changes, the prompt itself flags stale (independently of whether the rendered image is stale). Music prompts hash over `sequence.musicDesign` + analysis model.
5. No backfill — existing prompts are simply the row before the first new variant. Variant chain starts from the next save.

## Acceptance

- Editing a prompt produces a new variant row tagged `'user-edit'`; the previous AI-generated text is recoverable.
- Regenerating a prompt produces a `'regenerated'` row with a populated `input_hash`.
- A frame whose upstream context has changed shows a "prompt stale" indicator, separate from "image stale".

tombeckenham:
## Related Issue
Closes #620

## Summary
Implemented Stage 4 prompt versioning: new variants tables, input-hash columns, scoped helpers, wired all six write paths, and tests; pushed to open PR.

## Notes


tombeckenham (commented):
Round 1 review — Stage 4 prompt versioning.

## Summary
Two new variants tables, three input-hash columns, scoped helpers with revision history, and six write paths wired. Acceptance tests run against an in-memory libSQL DB with real migrations applied (good — no shallow mocks). No blocking issues. A few important notes below.

## Important

1. **Comment claims "single transaction" but implementation is sequential** — `src/lib/db/scoped/frame-prompt-variants.ts:1115-1117` and `src/lib/db/scoped/sequence-music-prompt-variants.ts:1115-1117` (paraphrasing line refs) document the operation as "in a single transaction", while the body (and the inline note immediately below) explains it's actually two sequential writes. The inline note is honest, but the leading docstring contradicts it. Consider reworking the docstring to say "appends a row and updates the cached pointer (sequentially today; see note)" so future readers don't rely on transactional semantics that don't exist. If a process crashes between the insert and the update, the cached `frames.imagePrompt`/`sequences.musicPrompt` will lag the variant chain — worth either an inline reconciliation pass on read, or a TODO that points to where atomicity will be added.

2. **`computeMotionPromptInputHash` never includes `elementBible`** — `src/lib/workflows/motion-prompt-scene-workflow.ts` does not destructure or pass `elementBible` to the hash, so the motion-prompt input surface always hashes `elementBible: null`. Visual prompt does include it. If element-bible changes are ever expected to invalidate motion prompts (e.g. a prop swap that affects camera language), staleness will not be flagged. Either drop `elementBible` from `PromptSceneContextHashInput` entirely (signal that motion intentionally ignores it) or thread it through the motion workflow like visual does. Today the schema permits it but the call site silently never sets it — confusing.

3. **`generateMusicFn` may write empty-string variants** — `src/functions/sequences.ts:395-403`: `prompt: data.prompt ?? sequence.musicPrompt ?? ''` and `tags: ... ?? ''`. The outer guard ensures *one* of `data.prompt`/`data.tags` is set, but if only `tags` is provided and `sequence.musicPrompt` is null, an empty-string `prompt` is persisted into the variants table as a `'user-edit'` row. The `prompt` column is `notNull` in the schema, so it'll succeed, but you'd be appending a meaningless empty prompt to history. Consider skipping the variant write when the prompt is empty, or require non-empty prompt at the validation layer.

4. **`PromptSceneContextHashInput` uses `unknown` for the entire upstream surface** — `src/lib/ai/input-hash.ts:234-249`: `scene`, `styleConfig`, `characterBible`, `locationBible`, `elementBible` are all `unknown`. This is inconsistent with the existing typed `*HashInput` types in the same file (e.g. `FrameImageHashInput` is concrete). Since `Scene`, style config, and bible types already exist, using them here would prevent silent shape drift between callers and the hash. Same for `MusicPromptInputHashInput.musicDesign`.

## Nits

- `frame-prompt-variants.ts` schema columns `components` and `parameters` are typed `$type<unknown>()` — a concrete type (matching the variants on `frames.metadata.prompts.visual`) would propagate through `FramePromptVariant` for free.
- The "Outstanding hardening" pointer in the docstring would be more useful as a `// TODO(#xxx)` linking the follow-up issue if one exists.
- `generateMusicPromptWorflow` (typo `Worflow` in the export) is pre-existing; not introduced here, just flagging.

## Verdict
No critical issues. Tests are solid and exercise the acceptance criteria from #620 directly. Important items above are worth addressing before merge but don't block correctness for the happy paths.

tombeckenham:
## Round 4 review-response

Round 4 review found no critical issues on the Stage 4 surface. All round 1 concerns were verified resolved by commits b3fd62e, 315b1327, 7cd398e (helpers' "single transaction" docstring corrected, `elementBible` threaded through `computeMotionPromptInputHash`, empty-prompt guard at `generateMusicFn`, `PromptSceneContextHashInput` typed concretely, variant `components`/`parameters` typed concretely).

This commit (50482329) addresses the three round 4 follow-ups:

1. **Test docstring corrected** — `frame-prompt-variants.test.ts:10-13` previously claimed the cached pointer is updated *atomically* by the helper, contradicting the helper's own docstring (which now correctly says "sequential, not transactional"). Reworded to match.

2. **Empty-prompt guard symmetry** — `music-prompt-workflow.ts` now throws on empty `musicDesignResult.prompt` before persisting a variant, matching the existing guard in `visual-prompt-scene-workflow.ts:79-83`. Closes the LLM-emits-empty-string parity gap.

3. **User-edit predicate extracted + tested** — the three-conjunct guard (`userEditedPrompt && prompt && prompt !== frame.imagePrompt`) was inline in both `image-workflow.ts` and `motion-workflow.ts`, with a comment warning that dropping any conjunct creates "phantom user edits" — but no test pinning that contract. Extracted to `src/lib/workflows/user-edit-predicate.ts` (`shouldRecordUserEdit`) with 8 unit tests covering all four cases: false flag, empty prompt, no-op edit, real edit (including first-edit when cached is null).

## Out-of-scope (recommend separate ticket)

Silent-failure review surfaced real issues in adjacent realtime/billing code that this PR did not introduce but did not fix either. Listing here so they aren't lost:
- `image-workflow.ts:139-156` — image generated but credits silently skipped on insufficient balance (`console.warn` only)
- `shot-variant-workflow.ts:264-272` — failure handler marks `thumbnailStatus: 'failed'` instead of the variant-image status
- `shot-variant-workflow.ts:295-297` — explicit `// Ignore emit errors` empty catch
- `motion-workflow.ts:151,379,444` — `void`-discarded emit promises bypass the surrounding try/catch
- `realtime/index.ts:251-281` — no-op channel stub silently drops events when an id is undefined

## Verification

- `bun typecheck` clean
- 174 tests pass across `src/lib/workflows/`, `src/lib/db/scoped/`, `src/lib/ai/__tests__/` (3 skipped, 0 failed)
- New `user-edit-predicate.test.ts`: 8/8 pass

PR #651, opened 2026-05-03
Stage 2: sheet variants for character / location / talent #618

### LINKED ISSUE #618: Stage 2: sheet variants for character / location / talent
Adds variants tables for sheet entities so divergence-on-completion can save a competing sheet without overwriting the live one. See [workflow-snapshots-and-content-hash-staleness](docs/architecture/workflow-snapshots-and-content-hash-staleness.md) § "Stage 2: sheet variants".

Depends on #617 (Stage 1 fully landed).

## Scope

1. New tables (Drizzle), each shaped like `frame_variants`: parent FK, model, URL/path, `input_hash`, `diverged_at`, `status`, `error`, `created_at`.
   - `character_sheet_variants` — parent: `characters`.
   - `location_sheet_variants` — covers both `locationSheets` and `locationLibrary` (parent FK type-tagged).
   - `talent_sheet_variants` — parent: `talent_sheets`.
2. Replace the Stage 1 "discard and re-queue" branch in `characterSheetWorkflow`, `locationSheetWorkflow`, and `libraryTalentSheetWorkflow` with: insert a row into the matching variants table, emit `stale:detected` with the new variant id.
3. UI: "alternate available" affordance on character/location/talent detail views, parallel to what frames already get. Wire to the same realtime event from #616.

## Trigger

Pick this up when real-world divergence on sheet workflows starts happening (recasts during generation are the obvious case) or when users start asking to compare sheet outputs across models.

## Acceptance

- New tables exist with the same shape as `frame_variants`.
- Sheet workflows route divergence to the variants tables instead of re-queuing.
- Detail views display the alternate-available affordance.

tombeckenham:
## Related Issue
Closes #618

> Stacked on `620-stage-4-prompt-versioning-and-variants`. Merge the parent PR first; GitHub will auto-retarget this one.

## Summary
Added character/location/talent sheet variants tables (mirroring frame_variants), scoped DB methods, divergence helper + tests, and stale:detected realtime event

## Notes


tombeckenham (commented):
## Review — Stage 2 sheet variants (round 1)

Scope is clean and additive: three new `*_sheet_variants` tables that mirror `frame_variants`, scoped CRUD wrappers, a pure `decideSheetDivergence` helper with unit tests, three `saveDivergent*` helpers, and a new `stale:detected` realtime event. Nothing existing is modified beyond schema/scoped/realtime registrations. Good test coverage on the pure decision function; helpers and tables are not yet wired to any workflow, which is consistent with the staged rollout.

### Important

- **`saveDivergentTalentSheet` conflates entity types in the realtime payload** — `src/lib/workflows/sheet-divergence.ts:949-954` emits `entityType: 'talent'` with `entityId: talentSheetId`. The realtime schema (`src/lib/realtime/index.ts:693-707`) defines `entityType: 'talent'` as a top-level talent entity, so consumers receiving `'talent' + <talentSheetId>` will look up a `talent` row and miss. A talent has many sheets (per the schema doc on `talent-sheet-variants.ts:359-363`), so the IDs are not interchangeable. Either add `'talent-sheet'` to the realtime `entityType` enum, or resolve the parent `talentId` from the talent sheet before emitting. The character/location helpers are fine because their `entityId` matches the declared `entityType`.

- **`saveDivergent*` helpers have no integration coverage** — only the pure `decideSheetDivergence` is tested. The save helpers wrap a DB insert + a realtime emit and are the surface that Stage 3 will depend on. A small test asserting the variant row gets `divergedAt`/`inputHash`/`status: 'completed'` populated and that `stale:detected` is emitted with the right shape would catch regressions cheaply. Acceptable to defer if Stage 3 will wire them with their own tests, but worth flagging now.

### Nits

- `src/lib/db/schema/{character,location,talent}-sheet-variants.ts`: the partial unique index is named `*_primary_key` (e.g. `character_sheet_variants_primary_key`), but the actual `PRIMARY KEY` is `id`. The "logical primary key" framing is fine in code, but a future reader debugging an integrity violation will find the name misleading. Consider `*_live_unique` or `*_undivergent_unique`.
- `inputHash` is nullable on all three tables, and the divergent unique index `(parent..., model, input_hash) WHERE diverged_at IS NOT NULL` would silently allow multiple `input_hash IS NULL` divergent rows in SQLite. The `saveDivergent*` helpers always supply `snapshotInputHash`, so this isn't reachable today, but a `CHECK (diverged_at IS NULL OR input_hash IS NOT NULL)` would make the invariant load-bearing.
- `drizzle/migrations/20260503053623_faulty_pestilence/migration.sql:67` is missing a trailing newline.
- The `STATUSES` literal-tuple-plus-`$type` pattern is duplicated verbatim across the three schema files. A shared `sheetVariantStatus` const would shrink the next-stage diff but is purely cosmetic.

### Verdict

No critical/blocking issues. The talent `entityType` mismatch is the main thing worth a follow-up before the consumer side lands.

PR #652, opened 2026-05-03
Stage 1 UI: frame staleness + divergence wiring #625

### LINKED ISSUE #625: Stage 1 UI: frame staleness + divergence wiring
Wires the Stage 1 UI primitives into the frame card and frame detail surfaces, per [staleness-and-divergence-ux.md](docs/architecture/staleness-and-divergence-ux.md) § "Staleness surfacing matrix" and § "Divergence resolution flow".

Supersedes the "wire one UI surface" bullet from #616 — that ticket should be rescoped backend-only (event schema + emission + isStale reader).

Depends on:
- #614 (input-hash schema columns)
- #615 (regenerateFramesWorkflow writes divergent rows into frame_variants)
- #616 backend half (`stale:detected` event schema + emission)
- #624 (UI primitives)

## Scope

1. **Corner indicator on frame card** — extend `src/components/scenes/scene-list-item.tsx` to render `<StalenessIndicator density="corner-dot">` or `<DivergentAlternateBanner density="corner-dot">` in the existing top-right slot (where Check / Loader2 / Skeleton render today, lines 58–81). Reads:
   - `frame.thumbnailInputHash` / `videoInputHash` / `audioInputHash` vs the recomputed hash from current upstream entities (via the scoped `isStale` reader from #614).
   - Any `frame_variants` row for this frame with `diverged_at IS NOT NULL` and `discarded_at IS NULL`.
2. **Inline banners in scenes-view right rail** — extend `src/components/scenes/scene-script-prompts.tsx`:
   - Above the image-prompt tab's action buttons (line ~688 area) when thumbnail is stale or has a divergent alternate.
   - Above the motion-prompt tab's action buttons when video is stale or has a divergent alternate.
3. **`useStaleDetected` hook** — `src/lib/realtime/use-stale-detected.ts`. Subscribes to `generation.stale:detected` on the current sequence channel, dispatches:
   - Sonner toast (debounced to one per sequence per 5s with a count).
   - TanStack Query invalidation for `frameKeys.list(sequenceId)` and the matching variants query.
4. **`<DivergenceCompareDialog>`** — `src/components/scenes/divergence-compare-dialog.tsx`. Two-column compare per spec (stage 1: upstream-entity-level diff only; field-level lands in stage 4). Calls Promote / Discard mutations from inside.
5. **Promote / Discard mutations** — extend `src/hooks/use-frames.ts` with `usePromoteVariantToPrimary` and `useDiscardVariant`. Server functions in `src/functions/frames.ts`. Scoped DB methods on `src/lib/db/scoped/frame-variants.ts` (`promote(...)`, `discard(...)`).
6. **Promote behaviour** — copies variant url/path into `frames.thumbnailUrl`/`thumbnailPath` (etc.), sets the matching `*_input_hash` column on `frames` to the variant's `input_hash`, soft-deletes the variant row, emits a synthetic `image:progress` (or matching) event with `status: completed`.
7. **Discard behaviour** — sets `discarded_at = now()` on the variant. Sonner toast with Undo action that clears `discarded_at`.

## Out of scope

- Sheet entity surfaces (character / location / talent) — that's Stage 2 UI.
- Sequence-level merged video / music — Stage 3 UI.
- Field-level diffing in compare dialog — Stage 4.
- Bulk "regenerate all stale" — deferred (spec § "Bulk operations").
- Prompt-stale indicators on prompt tabs — Stage 4 UI.

## Acceptance

End-to-end on a real sequence:
- Trigger `regenerateFramesWorkflow` on a frame; mutate the underlying character mid-flight; the workflow lands a divergent row in `frame_variants`; the realtime event fires; `<DivergentAlternateBanner>` appears in `scenes-view` without a manual refresh; the corner dot appears on `scene-list-item`.
- Click Compare → side-by-side dialog renders both images with the upstream-entity diff list.
- Click Promote → confirm dialog → primary `thumbnailUrl` swaps; `*_input_hash` updates; banner clears; corner dot clears.
- Click Discard → variant hides; sonner toast offers Undo; clicking Undo restores it.
- A frame whose upstream character or location has changed since generation (no in-flight workflow) shows `<StalenessIndicator>` in the matching tab and a corner dot on its card.

## Test coverage

- Unit tests on the promote/discard server functions (existing `__tests__/` pattern in `src/functions/`).
- Hook test for `useStaleDetected` covering the debounce + invalidation paths.
- Storybook stories for the dialog covering thumbnail / video / audio variants.

tombeckenham:
## Related Issue
Closes #625

> Stacked on `618-stage-2-sheet-variants-for-character-location-talent`. Merge the parent PR first; GitHub will auto-retarget this one.

## Summary
Implemented stage 1 staleness/divergence UI: discardedAt column + migration, scoped methods, promote/discard server fns + hooks, useStaleDetected realtime hook, DivergenceCompareDialog, scene-list-item corner dot, inline banners, scenes-view wiring with Undo toast; tests + story added; typecheck/lint clean; pushed.

## Notes

### DIFF
--- a/drizzle/migrations/20260503055203_sticky_xorn/migration.sql
+++ b/drizzle/migrations/20260503055203_sticky_xorn/migration.sql
@@ -1,0 +1,1 @@
+ALTER TABLE `frame_variants` ADD `discarded_at` integer;
--- a/drizzle/migrations/20260503055203_sticky_xorn/snapshot.json
+++ b/drizzle/migrations/20260503055203_sticky_xorn/snapshot.json
@@ -1,0 +1,6525 @@
+{
+  "version": "7",
+  "dialect": "sqlite",
+  "id": "9605b368-2d17-427f-a234-bd13adddf76a",
+  "prevIds": ["afc355d2-d8df-460a-a705-1e4363953231"],
+  "ddl": [
+    {
+      "name": "account",
+      "entityType": "tables"
+    },
+    {
+      "name": "audio",
+      "entityType": "tables"
+    },
+    {
+      "name": "character_sheet_variants",
+      "entityType": "tables"
+    },
+    {
+      "name": "characters",
+      "entityType": "tables"
+    },
+    {
+      "name": "credit_batches",
+      "entityType": "tables"
+    },
+    {
+      "name": "credits",
+      "entityType": "tables"
+    },
+    {
+      "name": "frame_prompt_variants",
+      "entityType": "tables"
+    },
+    {
+      "name": "frame_variants",
+      "entityType": "tables"
+    },
+    {
+      "name": "frames",
+      "entityType": "tables"
+    },
+    {
+      "name": "gift_token_redemptions",
+      "entityType": "tables"
+    },
+    {
+      "name": "gift_tokens",
+      "entityType": "tables"
+    },
+    {
+      "name": "location_library",
+      "entityType": "tables"
+    },
+    {
+      "name": "location_sheet_variants",
+      "entityType": "tables"
+    },
+    {
+      "name": "location_sheets",
+      "entityType": "tables"
+    },
+    {
+      "name": "passkey",
+      "entityType": "tables"
+    },
+    {
+      "name": "sequence_elements",
+      "entityType": "tables"
+    },
+    {
+      "name": "sequence_locations",
+      "entityType": "tables"
+    },
+    {
+      "name": "sequence_music_prompt_variants",
+      "entityType": "tables"
+    },
+    {
+      "name": "sequences",
+      "entityType": "tables"
+    },
+    {
+      "name": "session",
+      "entityType": "tables"
+    },
+    {
+      "name": "styles",
+      "entityType": "tables"
+    },
+    {
+      "name": "talent",
+      "entityType": "tables"
+    },
+    {
+      "name": "talent_media",
+      "entityType": "tables"
+    },
+    {
+      "name": "talent_sheet_variants",
+      "entityType": "tables"
+    },
+    {
+      "name": "talent_sheets",
+      "entityType": "tables"
+    },
+    {
+      "name": "team_api_keys",
+      "entityType": "tables"
+    },
+    {
+      "name": "team_billing_settings",
+      "entityType": "tables"
+    },
+    {
+      "name": "team_invitations",
+      "entityType": "tables"
+    },
+    {
+      "name": "team_members",
+      "entityType": "tables"
+    },
+    {
+      "name": "teams",
+      "entityType": "tables"
+    },
+    {
+      "name": "transactions",
+      "entityType": "tables"
+    },
+    {
+      "name": "user",
+      "entityType": "tables"
+    },
+    {
+      "name": "verification",
+      "entityType": "tables"
+    },
+    {
+      "name": "vfx",
+      "entityType": "tables"
+    },
+    {
+      "type": "text",
+      "notNull": false,
+      "autoincrement": false,
+      "default": null,
+      "generated": null,
+      "name": "id",
+      "entityType": "columns",
+      "table": "account"
+    },
+    {
+      "type": "text",
+      "notNull": true,
+      "autoincrement": false,
+      "default": null,
+      "generated": null,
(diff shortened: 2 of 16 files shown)


tombeckenham (commented):
Automated overnight review (round 1/5). Focused on the staleness/divergence UI wiring.

## Critical

### `SceneList` memo blocks re-render when divergent variants populate
`src/components/scenes/scene-list.tsx:222-273`

The `areEqual` comparator never checks `divergentVariants`, `staleThumbnailFrameIds`, `onCompareDivergent`, or `onRegenerateThumbnail`. After first paint with `divergentVariants === undefined`, when `useDivergentVariants` resolves (or `useStaleDetected` invalidates the cache after a realtime event), the parent re-renders with new `divergentVariants` but `frames` keeps the same array reference — line 251 short-circuits to `return true` and the memoized child is skipped. Net effect: the corner-dot indicator that is the headline UX of this PR will not appear until something else forces a rerender (frame status flip, selection change, etc.). Same applies when an alternate is promoted/discarded — the dot won't disappear.

Fix: add reference checks for `divergentVariants` and `staleThumbnailFrameIds` (and the new callbacks) to `areEqual` before the `frames === frames` early return.

## Important

### `promoteVariantFn` audio branch never marks audio completed
`src/functions/frames.ts:127-133`

The image branch sets `thumbnailStatus = 'completed'` and the video branch sets `videoStatus = 'completed'`, but the audio branch only writes `audioUrl`, `audioPath`, and `audioInputHash`. `frames.audioStatus` (`src/lib/db/schema/frames.ts:88`) keeps its prior value, which can leave a promoted audio alternate stuck at `pending`/`generating`/`failed`. Mirror the other two branches with `update.audioStatus = 'completed'` and clear `audioError` if applicable.

### `staleThumbnailFrameIds` declared but never wired
`src/components/scenes/scene-list.tsx:24` / `scenes-view.tsx:567-585`

`SceneList` consumes `staleThumbnailFrameIds` to drive the non-divergent staleness corner dot, but `scenes-view.tsx` never computes or passes it. The `isThumbnailStale` branch in `scene-list-item.tsx` is therefore dead in this PR. If staleness-without-alternate is intended for a follow-up stage, consider deleting the prop here so the contract isn't misleading; otherwise wire it up via the same realtime channel.

### `useStaleDetected` debounce timer leaks across unmount
`src/lib/realtime/use-stale-detected.ts:62-71`

The 5s `setTimeout` is stored on a ref but never cleared on unmount or when `sequenceId` changes — if the user navigates away within the debounce window, the toast fires against a stale sequence (and resets `count`/`timeout` on a no-longer-mounted ref, which is fine but the toast is the user-visible side effect). Add a `useEffect` cleanup that clears `debounceRef.current.timeout`.

### `divergence-compare-dialog.tsx` Discard button has no confirmation/Undo here
`src/components/scenes/divergence-compare-dialog.tsx:152-168`

Discard from the dialog routes through `handleDiscardWithUndo` in `scenes-view.tsx`, which does emit an Undo toast — good. But the dialog closes synchronously via `setCompareVariant(null)` immediately after dispatching the mutation (`scenes-view.tsx:660-663`), so if the mutation fails the user sees only the error toast with no path back to the dialog. Consider keeping the dialog open until `onSuccess` / `onError` resolve, mirroring how Promote already does (`scenes-view.tsx:223-225`).

## Nits

- `divergence-compare-dialog.tsx:69` uses `eslint-disable-next-line @next/next/no-img-element` but this is a TanStack Start project, not Next.js. The justifying comment is fine; just drop the irrelevant rule name (or switch to `unpic` per the project's image convention).
- `scenes-view.tsx:131` reads `useFramesBySequence`'s cache via `frameKeys.list(sequenceId)` for the refetchInterval predicate — pre-existing, not from this PR, just noting it lives next to the new wiring.
- `frame-variants.ts:247` `discard()` reuses the same column for "user discarded" and "promoted" (promote calls `discard(variant.id)` to soft-hide). Functionally equivalent for current listings, but a future audit/log surface would want to distinguish — worth a comment if not a separate column.

Tests look good — the partial-index split coverage (primary vs divergent uniqueness) and the new discard/undiscard/listDivergent tests are exactly the right shape.

PR #704, opened 2026-05-14
Regenerate Prompts #685

### LINKED ISSUE #685: Regenerate Prompts
### Description

Add the ability to regenerate either the scene image prompt or motion prompt

tombeckenham:
## Related Issue
Closes #685

## Summary
Implemented explicit Regenerate Prompt buttons + server-side force flag + DB write path for force-regen-same-hash; tests + typecheck green; pushed for CI.

## Notes

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "agent-harness": "bunx --bun github:tombeckenham/agent-harness",
     "dev": "bun run --parallel dev:db stripe:dev qstash:dev",
     "dev:db": "bun --bun --sequential db:migrate:local db:seed:local dev:e2e",
-    "dev:e2e": "bun --bun scripts/build-content-collections.ts && bun --bun vite dev",
+    "dev:e2e": "bun --bun scripts/build-content-collections.ts && bun --bun --preload ./scripts/bun-preload-idle-timeout.ts vite dev",
     "stripe:dev": "source .env.local 2>/dev/null; [ -z \"${STRIPE_SECRET_KEY}\" ] && echo 'Skipping Stripe (no STRIPE_SECRET_KEY)' && exit 0; stripe listen --forward-to localhost:3000/api/billing/webhook --api-key=${STRIPE_SECRET_KEY}",
     "qstash:dev": "docker run -p 8080:8080 public.ecr.aws/upstash/qstash:latest qstash dev",
     "build": "vite build",
--- a/scripts/bun-preload-idle-timeout.ts
+++ b/scripts/bun-preload-idle-timeout.ts
@@ -1,0 +1,31 @@
+/**
+ * Disable Bun.serve's 10s default idleTimeout for the vite dev server, which
+ * kills long-running streaming server-fn responses (structured-output LLM
+ * streams) mid-flight with "request timed out after 10 seconds" and a client
+ * "TypeError: network error".
+ *
+ * Bun's node:http shim caches its Bun.serve reference before vite plugins
+ * run, so patching from inside vite.config.ts is too late. Bun's --preload
+ * runs this file before any other module loads, which catches every Bun.serve
+ * call from anywhere in the process.
+ *
+ * Wired up via package.json: `bun --bun --preload ./scripts/bun-preload-idle-timeout.ts vite dev`.
+ * No effect on prod — the post-build Nitro patch in
+ * `scripts/patch-bun-idle-timeout.ts` handles that path.
+ */
+export {};
+
+type ServeFn = (opts: Record<string, unknown>) => unknown;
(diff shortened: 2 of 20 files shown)


tombeckenham (commented):
## PR review (round 1 of 5)

Automated review by overnight-harness agent. **Verdict: clean** — no blocking issues. The force-regen path is correctly gated, the new dedup-id helper is sound, and the DB-layer tests pin down the load-bearing invariants of the new branch (distinct row id, null `input_hash`, history ordering, cached hash still tracks live upstream). Per-tab busy indicator survives the mutation-variables shape change. Toast / aria-label / shadcn / Tailwind-layout-only conventions all met.

A few non-blocking items worth a human eye before merge.

### Important

**1. Misleading comment about why `user-edit`/`restored` can't reach the new branch.** `src/lib/db/scoped/frame-prompt-variants.ts:157-160` says they *\"never reach this branch — they don't carry a non-null \`inputHash\` here.\"* That justification is wrong — the type at lines 56-73 explicitly allows `user-edit` and `restored` to carry a non-null hash, and `src/lib/workflows/image-workflow.ts:121-140`, `src/lib/workflows/motion-workflow.ts:166-185`, and `restoreFramePromptVariantFn` (`src/functions/prompt-variants.ts:143-153`) all pass a non-null hash in the happy path. What actually keeps them out is the `source === 'ai-generated' || source === 'regenerated'` predicate at line 155. Risk: a future maintainer reads the comment, concludes the source check is redundant, and removes it — silently changing behaviour. Suggest rewording to make the source-filter the load-bearing guard explicitly.

**2. Pre-existing user-edit-at-same-hash silent text loss is now explicitly carved out.** When a user edit produces text whose live upstream hash collides with an existing AI/regenerated row (`src/lib/db/scoped/frame-prompt-variants.ts:176-178`), the `onConflictDoNothing()` no-ops, the new branch's source filter excludes `user-edit`, and the code returns `existing` (the OLD AI text) — but the cached `frames.imagePrompt` column is then updated to the user's new text at lines 185-192. Net: card shows the user's edit, history shows only the AI text, and a later \"restore latest user-edit\" would restore the AI text. This pre-dates the PR, but the new docstring at lines 41-44 (*\"User-edits also carry the live hash captured at edit time so staleness detection keeps working after a hand-typed prompt\"*) reads as if the user-edit row is being persisted — it isn't, in the colliding case. Either extend the new branch to include `'user-edit'` when text differs, or log a warning in the `else` so the divergence is observable.

**3. Non-transactional second write widens the inconsistency window.** The variant insert and the `UPDATE frames` at `src/lib/db/scoped/frame-prompt-variants.ts:185-192` aren't wrapped in a transaction (acknowledged by the file's top doc). If the UPDATE fails after the force-regen INSERT succeeded, a null-hash row exists in history but `frames.imagePrompt` still points at the old text. Because the new row has `input_hash = NULL`, the partial-unique index can no longer dedupe — every user retry click writes another null-hash history row. Pre-existing structural issue, but the force-regen path widens its blast radius. Worth at least a `logError` when the second write throws so this state is visible.

**4. No test for the source-filter guard.** Tests cover the positive force-regen case (`source: 'regenerated'`, same hash, different text) and idempotent same-text retry, but there's no test asserting `source: 'user-edit'` at the same hash with different text does *not* take the new fallback branch. A test pinning the intended contract for that case would catch a regression that broadens the predicate (and would force the team to decide whether the silent-loss behaviour above is intended).

### Nits

- `src/lib/db/scoped/frame-prompt-variants.ts:44` — *\"the helper may downgrade the persisted hash to null on the force-regen fallback path\"* reads as if the cached `*_prompt_input_hash` column is nulled. It isn't — only the inserted variant row carries `input_hash = NULL`; the cached column always gets `nextHash`. Tighten the wording.
- `src/lib/db/scoped/frame-prompt-variants.ts:106` — *\"The new completion's hash matches an existing row\"* — disambiguate as \"upstream-input hash\" to avoid confusion with the `simpleHash(text)` used elsewhere.
- `src/functions/prompt-variants.ts:267` — `crypto.randomUUID()` works in Bun + Cloudflare Workers, but the codebase otherwise uses `generateId()` (ULID) from `src/lib/db/id.ts`. The `Date.now()` prefix is redundant given `randomUUID()` is already collision-free; not a bug, just consistency.
- `src/lib/db/scoped/frame-prompt-variants.ts:181-183` — the \"Failed to insert frame prompt variant\" throw has no IDs in the message; if the lookup-after-conflict race ever fires, it will be very hard to diagnose. Consider including `frameId` / `promptType` / `nextHash` in the error or routing through `logError` with an error id.
- `src/functions/prompt-variants.ts:259` — *\"unique dedup ID per click\"* is technically *\"per server-fn invocation\"*; transport-layer retries would still bypass dedup. Worth a small wording tweak.

PR #705, opened 2026-05-14
Select video and / or music model when generating #686

### LINKED ISSUE #686: Select video and / or music model when generating
### Description

When the user is ready to generate video, there is no way to change the video model they use. This is only changable right at the beginning when generating the sequence. The user should be able to set the sequence video model at generation time. The same goes for music

tombeckenham:
## Related Issue
Closes #686

> Stacked on `685-regenerate-prompts`. Merge the parent PR first; GitHub will auto-retarget this one.

## Summary
Added motion + music model selectors to the batch-generate footer (desktop and mobile) and plumbed both models through to batchGenerateMotionFn.

## Notes

### DIFF
--- a/src/components/scenes/mobile-scene-drawer.tsx
+++ b/src/components/scenes/mobile-scene-drawer.tsx
@@ -1,3 +1,5 @@
+import { MotionModelSelector } from '@/components/model/motion-model-selector';
+import { MusicModelSelector } from '@/components/model/music-model-selector';
 import { Button } from '@/components/ui/button';
 import { Checkbox } from '@/components/ui/checkbox';
 import { ScrollArea } from '@/components/ui/scroll-area';
@@ -8,11 +10,18 @@ import {
   SheetHeader,
   SheetTitle,
 } from '@/components/ui/sheet';
+import {
+  DEFAULT_MUSIC_MODEL,
+  DEFAULT_VIDEO_MODEL,
+  type AudioModel,
+  type ImageToVideoModel,
+} from '@/lib/ai/models';
 import type { AspectRatio } from '@/lib/constants/aspect-ratios';
 import { cn } from '@/lib/utils';
 import type { Frame } from '@/types/database';
 import { ChevronUp, Loader2, Video } from 'lucide-react';
-import { useMemo, useState } from 'react';
+import { useMemo, useRef, useState } from 'react';
+import type { BatchGenerateMotionArgs } from './scene-list';
 import { SceneListItem } from './scene-list-item';
 import { SceneThumbnail } from './scene-thumbnail';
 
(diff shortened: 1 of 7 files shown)


tombeckenham (commented):
## Automated review — round 1

Scope: 4 files, mostly UI plumbing for motion/music model selection at the batch-generate footer plus a server-fn schema extension. No critical issues found; a few important refinements worth considering before merge.

### Important

**1. Discriminated-union the `BatchGenerateMotionArgs` invariant** — `src/components/scenes/scene-list.tsx:15-19`
`musicModel` is always required even when `includeMusic: false`, and `scenes-view.tsx:559` already encodes the real shape (`musicModel: includeMusic ? musicModel : undefined`). Modeling it as a union makes the conditional disappear and prevents a stale model being silently dropped:
```ts
type BatchGenerateMotionArgs =
  | { includeMusic: false; motionModel: ImageToVideoModel }
  | { includeMusic: true; motionModel: ImageToVideoModel; musicModel: AudioModel };
```
This would also align with the Zod schema, where `musicModel` is already `.optional()`.

**2. Credit preflight may use a different model than per-frame generation** — `src/functions/motion-functions.ts:152` vs `:197-200`
`requireCredits` is computed with `batchModel = data.model ?? DEFAULT_VIDEO_MODEL`, but per-frame the handler re-resolves via `safeImageToVideoModel(data.model || frame.motionModel || sequence.videoModel, …)`. When `data.model` is omitted (schema marks it optional), per-frame fallback can land on a more-expensive model than the preflight charged. The new UI always supplies `model`, but the schema doesn't enforce it. Either make `model` required in `batchGenerateMotionInputSchema`, or compute the preflight by summing per-frame resolved models.

**3. Silent `includeMusic` downgrade** — `src/functions/motion-functions.ts:166-167`
If the client posts `includeMusic: true` + `musicModel: 'X'` while `sequence.musicStatus === 'generating'`, the server flips `includeMusic` to false and returns it that way, but the UI has already shown the user "generate with model X". Per CLAUDE.md ("Don't create fallback code — display errors to user"), consider throwing instead of silently downgrading, so the user gets a clear "music is already generating" message.

**4. Cross-component type export coupling** — `src/components/scenes/scene-list.tsx:15`
`BatchGenerateMotionArgs` is exported from `scene-list.tsx` and re-imported by `mobile-scene-drawer.tsx` (its sibling) and `scenes-view.tsx`. Drawer importing a type from list is an unnecessary coupling. Move to `src/components/scenes/types.ts` (or `src/lib/...`).

**5. Test gap for `batchGenerateMotionFn`** — `src/functions/motion-functions.ts:133`
Pre-existing — not introduced by this PR — but the new `musicModel` forwarding contract (only when `includeMusic`) and the existing music-prompt/tags missing-throw are uncovered. A single Bun test mocking `scopedDb`, `triggerWorkflow`, and `requireCredits` would cheaply guard against contract drift. Worth a follow-up issue rather than blocking here.

### Nit

- **Prop→state sync pattern duplication** — `scene-list.tsx:90-102` and `mobile-scene-drawer.tsx:77-89` use the "adjust state during render" trick to mirror prop changes. It works under StrictMode, but two simpler equivalents exist: (a) `key={initialMotionModel}` on the footer container, or (b) treat `motionModel` as fully derived (`userOverride ?? initialMotionModel ?? DEFAULT_VIDEO_MODEL`) with `userOverride` as the only state. Reduces cognitive load and removes the duplication.
- **WHAT-not-WHY JSDoc** — the `initialMotionModel`/`initialMusicModel`/`styleCategory` doc comments in both `scene-list.tsx` and `mobile-scene-drawer.tsx` mostly restate the prop names. Per CLAUDE.md "Default to writing no comments… Don't explain WHAT the code does", these can be trimmed. The `// Sync local selection when…` comment in `scene-list.tsx:84` is a textbook WHY comment and should stay.
- **Inconsistent `includeMusic` default** — `false` in mobile drawer vs `true` in desktop list. Likely intentional (mobile = data-conscious) but worth a one-line comment to prevent a future "fix".
- **`z.enum(Object.keys(AUDIO_MODELS) as [keyof typeof AUDIO_MODELS])`** — consistent with the existing pattern in `frame.schemas.ts`. A small `enumFromRecord()` helper in `models.ts` would centralize the unsafe assertion across all four call sites, but that's out of scope here.

### Verdict
No blocking issues. Nice, tight PR — the type-safe boundary via Zod, the `safeAudioModel`/`safeImageToVideoModel` indirection, and the `music.model` plumbing into `motion-batch-workflow` all line up cleanly. The discriminated-union refactor (#1) and the credit preflight consistency (#2) are the two follow-ups most worth a fix before merge.

PR #706, opened 2026-05-14
Turn off music and sfx / dialogue #687

### LINKED ISSUE #687: Turn off music and sfx / dialogue
### Description

- There is no way to turn off sfx / dialogue when generating video. Add a checkbox to toggle (on by default)
- This should be on the scene level, and when batch generating
- Once music is generated, there is no way to turn it off. The music tab should have a checkbox to include music

tombeckenham:
## Related Issue
Closes #687

## Summary
- **Include SFX & dialogue** checkbox at scene, batch, and mobile-drawer levels for audio-capable models (kling v3, veo3). Default on. Wires through to the model's `generate_audio` field and adjusts cost estimation.
- **Include music in merged video** checkbox in the music tab. Routes the Merge with Video CTA through a new `skipAudioMux` flag so the merge-video workflow bypasses the audio mux when unchecked.
- UX fix: hide the batch-generate sticky footer while a motion job is in progress (was misleading — checkboxes implied they still influenced the run).
- UX fix: render the Compare/Promote/Discard row inline below the divergent-alternate banner text (was tucked into the corner via absolute positioning, hidden in the narrow scenes motion column).

## Test plan
- [ ] Open a sequence with a kling v3 or veo3 frame → "Include SFX & dialogue" appears on scene panel, batch footer, and mobile drawer, default checked
- [ ] Generate a single frame with the box **unchecked** → resulting motion has no audio track, cost estimate reflects the cheaper run
- [ ] Batch generate with mixed audio-capable + non-audio-capable models → checkbox only affects the audio-capable ones
- [ ] With music already generated, uncheck "Include music in merged video" on the Music tab and trigger Merge with Video → resulting mp4 has motion audio only, no music track
- [ ] While a batch motion job is running, the sticky footer is hidden; MotionProgressBanner still visible
- [ ] On a frame with a divergent alternate, Compare/Promote/Discard sits inline below the banner text

### DIFF
--- a/src/components/music/music-view.tsx
+++ b/src/components/music/music-view.tsx
@@ -2,6 +2,7 @@ import { MusicModelSelector } from '@/components/model/music-model-selector';
 import { PromptHistorySheet } from '@/components/prompts/prompt-history-sheet';
 import { StalenessIndicator } from '@/components/staleness/staleness-indicator';
 import { Button } from '@/components/ui/button';
+import { Checkbox } from '@/components/ui/checkbox';
 import { Input } from '@/components/ui/input';
 import { Label } from '@/components/ui/label';
 import { Skeleton } from '@/components/ui/skeleton';
@@ -31,12 +32,17 @@ type GenerateMusicArgs = {
   duration?: number;
 };
 
+type MergeVideoAndMusicArgs = {
+  /** When `false`, restitch frame videos but skip the music mux. */
+  includeMusic: boolean;
+};
+
 type MusicViewProps = {
   sequence: Sequence;
   videoDuration?: number;
   onGenerateMusic: (args?: GenerateMusicArgs) => void;
   isGeneratingMusic: boolean;
-  onMergeVideoAndMusic: () => void;
+  onMergeVideoAndMusic: (args: MergeVideoAndMusicArgs) => void;
   isMergingVideoAndMusic: boolean;
   /** Banner rendered above the audio player while `musicStatus === 'completed'`. */
   divergentBanner?: React.ReactNode;
@@ -160,6 +166,7 @@ export const MusicView: React.FC<MusicViewProps> = ({
     () => videoDuration
   );
   const [historyOpen, setHistoryOpen] = useState(false);
+  const [includeMusicInMerge, setIncludeMusicInMerge] = useState(true);
 
   // Resync the textarea when the source-of-truth musicPrompt changes from
   // outside (regenerate, restore, realtime update). Without this, a successful
@@ -254,6 +261,21 @@ export const MusicView: React.FC<MusicViewProps> = ({
         <ReadOnlyField label="Prompt" value={musicPrompt ?? 'Missing prompt'} />
         <ReadOnlyField label="Tags" value={musicTags ?? 'Missing tags'} />
 
+        <label
+          htmlFor="merge-include-music"
+          className="flex items-center gap-2 self-center text-sm text-muted-foreground"
+        >
+          <Checkbox
+            id="merge-include-music"
+            checked={includeMusicInMerge}
+            onCheckedChange={(checked) =>
+              setIncludeMusicInMerge(checked === true)
+            }
+            disabled={isMerging}
+          />
+          <span>Include music in merged video</span>
+        </label>
+
         <div className="flex justify-center gap-3">
           {historyButton}
           <LoadingButton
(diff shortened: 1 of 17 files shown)


tombeckenham (commented):
## Automated PR Review (Round 1)

Reviewed the SFX/dialogue + include-music toggles end-to-end. The core plumbing of `generateAudio` (UI → server fn → workflow → `buildModelInput`) and the `skipAudioMux` short-circuit are correct, validated, and respect the `videoModelSupportsAudio` guard at every UI submission site. No blocking issues — `clean` verdict.

A few things worth a second pass before merge:

### Important

1. **`mergeVideoAndMusic` mutation has no `onError` rollback** — `src/routes/_protected/sequences/\$id/music.tsx:170-185`. The optimistic update writes `mergedVideoStatus: 'merging'` in `onMutate`, but no `onError` rolls it back or surfaces a toast. New error paths reachable via this PR (`includeMusic=true && !musicUrl` race, the existing `frames.length===0`, `incompleteCount>0`, insufficient-credits throws in `sequences.ts:482-503`) will leave the music tab wedged on "Merging…" with no user-visible error. The sibling `regenerateMusicPrompt` mutation at lines 210-228 already shows the correct pattern (`onError` + `invalidateQueries` + toast). Per CLAUDE.md "Don't create fallback code — display errors to user."

2. **Pre-flight `estimateVideoCost` ignores `generateAudio`** — `src/functions/motion-functions.ts:80` and `:160`. `estimateVideoCost(model, duration)` defaults `audioEnabled: videoModelSupportsAudio(model)` in `src/lib/billing/cost-estimation.ts:62`, so pre-flight always treats audio-capable models as audio-on. A user on kling-v3/veo3 who unchecked SFX/dialogue can be pre-flight-blocked for inflated credit cost they wouldn't actually incur (real deduction in `motion-generation.ts:181-182` is correct). Suggest forwarding `audioEnabled: data.generateAudio !== false && videoModelSupportsAudio(model)` in both call sites.

3. **`build-model-input.ts` implicitly relies on Zod's strip-unknown-keys default** — `src/lib/motion/build-model-input.ts:49-54`. The comment says non-audio models "strip it during apiSchema.parse" — that's true today because the generated `zod.gen.ts` uses plain `z.object()`. If a future codegen run flips to `.strict()` (e.g., via `@hey-api/openapi-ts` flag change), every motion job submitted to Minimax/etc. will start throwing `ZodError` on unknown key `generate_audio`. Cheap to make the contract explicit: guard the spread with `videoModelSupportsAudio(modelKey)`. Add a `buildModelInput('minimax_hailuo_02', { generateAudio: false })` test to lock the contract.

4. **Test coverage gaps on new critical-path logic**:
   - `calculateMotionMetadata` audio derivation in `src/lib/motion/motion-generation.ts:181-182` is untested. `calculateVideoCost` itself is well covered in `src/lib/ai/fal-cost.test.ts`, but the wiring that decides what to pass as `audioEnabled` isn't. A future refactor of `!== false` to `!!options.generateAudio` would silently flip billing defaults. Natural home: existing `src/lib/motion/motion-generation.test.ts`.
   - `chainAudioMux` `skipAudioMux` short-circuit in `src/lib/workflows/merge-video-workflow.ts:50-55` has no test — load-bearing behavioral change with no regression guard.
   - `mergeVideoAndMusicFn` validation change in `src/functions/sequences.ts:479-483` — `data.includeMusic !== false` is identical-looking to `!== undefined`; only a handler test pins which one we meant.

### Nits

- `generateAudio` UI state isn't reset when toggling from an audio-capable model to a non-audio model and back (scene-list, mobile-scene-drawer, scene-script-prompts). Submission paths coerce correctly, so no functional bug — but the checkbox snaps back to whatever the user last set.
- Only 2 of 5 audio-capable models (kling_v3_pro, veo3_1) are covered in `build-model-input.test.ts`. `seedance_v2`, `seedance_v1_5_pro`, `ltx_2_3_pro` also expose `generate_audio` per `zod.gen.ts`.

### Confirmed correct

- `mergeVideoAndMusicFn` correctly only throws the music-required check when `includeMusic !== false`; default preserves legacy behavior.
- `chainAudioMux` early-return is logged with sequence ID for debuggability — not silent.
- `motionBatchWorkflow` passing `frame.generateAudio` (boolean | undefined) is type-safe with `MotionWorkflowInput.generateAudio`.
- UI gates `videoModelSupportsAudio` consistently at all submission sites.
- No `any` / `unknown` / `!` violations introduced.

---
🤖 Round 1/5 — overnight harness review

tombeckenham (commented):
Approved

PR #707, opened 2026-05-14
Edit scene duration #688

### LINKED ISSUE #688: Edit scene duration
### Description

Add the ability to edit scene duration - which would cause motion to be stale.

tombeckenham:
## Related Issue
Closes #688

> Stacked on `687-turn-off-music-and-sfx-dialogue`. Merge the parent PR first; GitHub will auto-retarget this one.

## Summary
Implemented editable scene duration in script tab (frames.durationMs + metadata.durationSeconds), fixed single-frame motion regen to honor stored duration; tests + lint green, pushed for PR.

## Notes


tombeckenham (commented):
## Automated review — round 1

Two files in scope; functionality works end-to-end. No blocking issues, but a few **Important** items worth addressing before this lands (or as a follow-up issue if you want to keep the PR scope tight).

### Important

**1. Pre-flight credit check uses unsnapped duration** — `src/functions/motion-functions.ts:88-92`

The new fallback chain pipes a stored `durationMs` (UI allows up to 60s) straight into `requireCredits(scopedDb, estimateVideoCost(model, duration), …)`. Most motion models snap to 5/10s inside `calculateMotionMetadata`, and the actual deduction in `motion-workflow.ts` uses that snapped duration. Net effect: a user with enough credits for the snapped 10s can be rejected at pre-flight with a cost estimated for 60s.

Note that `batchGenerateMotionFn` directly below (line 165) already snaps before `requireCredits`:
```ts
const batchDuration = snapDuration(data.duration, batchModel);
```
The single-frame path should do the same. Cheap fix; also resolves the inconsistency between the two functions.

**2. UI ceiling (60s) vs schema/model ceiling (~10s) silent overwrite** — `src/components/scenes/scene-script-prompts.tsx:862-863` vs `src/lib/schemas/frame.schemas.ts:85`

`generateMotionSchema.duration` is `z.number().min(1).max(10)`, and individual fal.ai models cap even tighter. The UI happily accepts 1–60. A user who types 60 will (a) pass the cost gate at the wrong rate (see #1), (b) get a video that's at most ~10s, and (c) see `frame.durationMs` overwritten back to the snapped value by `motion-workflow.ts:410`. No toast, no banner — the typed value silently disappears.

Suggestions, in order of effort:
- Clamp the input `max` per the currently-selected motion model (you already pass `selectedMotionModel` around in this component).
- Or add a notice next to the input: "Most motion models clamp to ≤10s — longer values will be shortened."
- Or align the UI to the schema's `max(10)`.

**3. Misleading validation copy for non-numeric input** — `src/components/scenes/scene-script-prompts.tsx:798-803, 872-876`

`Number('3,5')` → `NaN`, which the validity check catches via `Number.isFinite`. But the error reads "Duration must be between 1 and 60 seconds." — confusing when the user just pasted `3,5` and sees a "range" complaint for a value they consider valid. Consider distinguishing the two cases:

```ts
const errorMessage = !Number.isFinite(parsedDuration)
  ? 'Enter a number using a period as the decimal separator.'
  : 'Duration must be between 1 and 60 seconds.';
```

**4. Error message not linked to the input** — `src/components/scenes/scene-script-prompts.tsx:858-876`

`aria-invalid` is set correctly, but the inline error `<p>` has no id and the input has no `aria-describedby`. Screen readers won't announce the error when the field gets focus. Add `id="scene-duration-error"` on the `<p>` and `aria-describedby={!isDurationValid ? 'scene-duration-error' : undefined}` on the `Input`.

**5. No tests for the new duration-resolution branch** — `src/functions/motion-functions.ts:84-88`

This branch directly drives both **billed cost** (line 90) and the duration submitted to fal.ai, with a three-tier fallback (`durationMs/1000` → `metadata.metadata.durationSeconds` → `snapDuration(undefined, model)`). There is currently no test file for `motion-functions.ts`. Per `CLAUDE.md` "cover critical paths only", this qualifies. Consider extracting the resolution to a pure helper (e.g. `src/lib/motion/resolve-frame-duration.ts`) and adding a colocated `.test.ts` covering: explicit `data.duration` wins; positive `durationMs` → seconds; missing/zero `durationMs` → metadata fallback; missing metadata → `snapDuration` default; parity with `batchGenerateMotionFn`.

### Nit

- **`frame.durationMs ?` truthy check treats 0 as unset** — `src/functions/motion-functions.ts:84`. Inconsistent with the script tab's explicit `frame.durationMs > 0` check at `scene-script-prompts.tsx:782`. `updateFrameSchema.durationMs.min(1)` makes 0 unreachable today, but mirroring the explicit comparison would make intent legible.
- **Inconsistent "no duration" default** — `motion-functions.ts:88` falls back to `snapDuration(undefined, model)` while `motion-functions.ts:222-223` falls back to literal `3`. Pre-existing on the batch path, but the new code makes the divergence more visible. Worth filing a follow-up to pick one.
- **Redundant `?? undefined`** — `scene-script-prompts.tsx:784`. `??` on an already-`undefined` value is a no-op; drop it.
- **Dirty check is string-equality** — `scene-script-prompts.tsx:795-797`. Typing `'3.0'` against saved `'3'` reports dirty and Save activates, persisting the same value. Cosmetic. Consider comparing parsed numbers: `parsedDuration !== savedDurationSeconds`.
- **"Saving will mark prompts as stale" copy** — `scene-script-prompts.tsx:910-914`. Accurate (durationSeconds is part of the hashed scene context, confirmed in `input-hash.ts:sceneInputContext`), so the message stands; no change.

### Positive

- The `metadata.metadata ?? { title: '', … }` fallback at lines 282-289 is safe (`??` only fires on null/undefined; populated metadata is preserved verbatim) and writes both `durationMs` and `metadata.metadata.durationSeconds` for display/downstream consistency.
- State reset on frame switch via `prevScriptFrameIdRef` at line 637-641 follows the existing pattern in this component.
- Save mutation correctly invalidates `frameKeys.detail` / `frameKeys.list` / `frameStalenessKey` / `sequenceKeys.detail` so prompt-staleness banners light up after duration-only edits.
- Server middleware (`frameAccessMiddleware`) and `updateFrameSchema` already accept `durationMs`, so no schema changes needed.

### Verdict

No correctness or security blockers. The motion-functions credit-gate snap (#1) is mechanical and aligns this path with the existing batch path; everything else is shipable as-is with follow-up issues.

PR #708, opened 2026-05-14
Replace elements #689

### LINKED ISSUE #689: Replace elements
### Description

Add the ability to replace an element - like we have for cast members and locations. With elements specifically, the image frame should be _edited_ to replace the element not regenerated.

tombeckenham:
## Related Issue
Closes #689

> Stacked on `688-edit-scene-duration`. Merge the parent PR first; GitHub will auto-retarget this one.

## Summary
Implemented replace-element flow: per-element file picker + confirm dialog, server fn + workflow that edits affected frames in place via image-workflow's edit endpoint.

## Notes

### DIFF
--- a/e2e/fixtures/recorded/openrouter/element-vision/openai-2026-05-15T01-25-41-410Z-0623a6e5.json
+++ b/e2e/fixtures/recorded/openrouter/element-vision/openai-2026-05-15T01-25-41-410Z-0623a6e5.json
@@ -2,13 +2,13 @@
   "fixtures": [
     {
       "match": {
-        "userMessage": "Element token: BROADCAST_MIC\nUploaded filename: broadcast-mic.jpg\n\nDescribe the element in the image below.",
+        "userMessage": "Uploaded filename (hint only — may be meaningless): broadcast-mic.jpg\n\nDescribe the element in the image below and suggest a token.",
         "model": "anthropic/claude-sonnet-4.6",
         "turnIndex": 0,
         "hasToolResult": false
       },
       "response": {
-        "content": "{\"description\": \"A large-diaphragm condenser studio microphone with a cylindrical champagne/satin-nickel metal body. The upper capsule section features a fine mesh grille with a curved top. The body bears engraved text reading 'PHANTOM classic' and a serial number 'Nr. 1307', along with a circular logo emblem. The microphone is mounted in a silver spider-style shock mount with elastic bands. A circular black pop filter with dual-layer mesh is attached via a gooseneck arm to the same stand. The overall finish is matte champagne silver with black hardware accents.\", \"consistencyTag\": \"champagne-condenser-studio-microphone\"}"
+        "content": "{\"description\": \"A large-diaphragm condenser studio microphone with a cylindrical champagne/satin-nickel metal body. The upper capsule section features a fine mesh grille with a curved top. The body bears engraved text reading 'PHANTOM classic' and a serial number 'Nr. 1307', along with a circular logo emblem. The microphone is mounted in a silver spider-style shock mount with elastic bands. A circular black pop filter with dual-layer mesh is attached via a gooseneck arm to the same stand. The overall finish is matte champagne silver with black hardware accents.\", \"consistencyTag\": \"champagne-condenser-studio-microphone\", \"suggestedToken\": \"BROADCAST_MIC\"}"
       },
       "metadata": {
         "systemHash": "7547c536"
--- a/e2e/mocks/aimock-server.ts
+++ b/e2e/mocks/aimock-server.ts
@@ -63,7 +63,7 @@ const STAGE_PREFIXES: ReadonlyArray<readonly [string, string]> = [
   ['Generate the visual prompt for the starting frame', 'visual-prompts'],
   ['Generate the motion prompt for this scene', 'motion-prompts'],
   ['Classify music design for each scene', 'music-design'],
-  ['Element token:', 'element-vision'],
+  ['Uploaded filename (hint only', 'element-vision'],
 ];
 
 // Diagnostic dump for unmatched requests — written on shutdown so we can diff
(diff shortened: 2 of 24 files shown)


tombeckenham (commented):
## Automated review (round 1/5)

Three reviewers ran in parallel: code-reviewer, silent-failure-hunter, pr-test-analyzer. The PR is well-structured and matches existing patterns (server-handler/DB-only-in-handlers, scoped workflow, kebab-case files, named exports, no `useState`+`useEffect` for data, shadcn/ui base components, no `any`/`!`/`unknown`, `type` over `interface`). The `previousDescription` capture before the description-nulling update is correct. Reference image roles (`primary`/`element`) match `ReferenceImageDescription`. `matchElementsToScene` (the matching logic the new code depends on) is already covered in `scene-matching.test.ts`. No tests is consistent with how peer workflows (`recast-character-workflow.ts`, `recast-location-workflow.ts`, `regenerate-frames-workflow.ts`) are structured — workflow-shell tests are not the project's pattern.

### Critical

**1. Element stuck in `analyzing` forever if `triggerWorkflow` throws** — `src/functions/sequence-elements.ts:289-330`

The handler updates the element row to `visionStatus: 'analyzing'` (line 289), emits `:start` (line 309), then calls `triggerWorkflow` (line 326). If the trigger fails (QStash 5xx, network blip, label rejection), the element row remains in `analyzing` indefinitely. The mutation toasts \"Failed to replace element\" but the element card keeps spinning (`isReplacing` is true while `element.visionStatus === 'analyzing'`, see `element-card.tsx:71-72`). The user can only recover by deleting the element or uploading again. Wrap the trigger in a try/catch: on throw, restore vision status (`updateVisionStatus(elementId, 'failed', error.message)`) and emit `replace-element:failed`.

### Important

**2. Hard-throw aborts entire workflow if any affected frame has been deleted mid-flight** — `src/lib/workflows/replace-element-workflow.ts:131-138`

`affectedFrameIds` is captured in the server fn and travels through QStash; by execution time, one of those frames may have been deleted. Throwing aborts the workflow — no frames get edited and the element is marked vision-failed. The sibling `regenerate-frames-workflow.ts` handles this via per-frame `skipped-deleted` outcomes (see lines 172-207 there). Filter out missing frames and treat them as skipped.

**3. Duplicate `:start` emit and orphaned `:start` on zero-frames early return** — `src/functions/sequence-elements.ts:309-312` + `src/lib/workflows/replace-element-workflow.ts:101-117`

The server fn emits `replace-element:start`, then the workflow re-emits `:start` from its `emit-start` step. The recast pattern (`regenerate-frames-snapshot.ts:emitRecastEvent`) emits start/complete/failed exclusively from the workflow — never from the server fn. Worse, when `affectedFrameIds.length === 0` the workflow returns at line 110 *before* reaching the workflow-side `:start` emit, but the server fn already emitted `:start`. Result: a `:start` with no matching `:complete` or `:failed`, which will hang any future UI subscriber. Today no client subscribes to these events, so the user-visible impact is nil — but the lifecycle inconsistency will trip whoever wires the UI next. Recommend: remove the server-fn emit; emit `:complete` even on the zero-frames path.

**4. All-frames-fail silently reports `:complete` with zero successes** — `src/lib/workflows/replace-element-workflow.ts:145-244`

If every per-frame `context.invoke('image', …)` fails, `successCount === 0`, the workflow returns normally and emits `:complete`, the element stays `completed`, and the UI shows green even though no frame was actually edited. This mirrors `regenerate-frames-workflow.ts` behavior so it's a pre-existing pattern, but consider throwing when `failedCount === results.length && results.length > 0` so the failureFunction fires and the UI surfaces a real failure.

**5. N+1 frame-count queries on the elements grid** — `src/components/element/element-card.tsx:32` rendered by `src/components/element/elements-view.tsx:109-111`

`ElementCard` calls `useFrameIdsForElement(sequenceId, element.id)` for every element. Each call hits `getFrameIdsForElement` in the scoped DB, which loads ALL frames (with full JSONB metadata) for the sequence and filters in JS. With N elements, that's N full-frame scans on every elements-view mount. The peer hooks `useFrameIdsForCharacter`/`useFrameIdsForLocation` are only used in single-entity detail views, so they don't have this cost. Consider a batched `getFrameCountsForAllElements(sequenceId)` that loads frames once and matches all elements in a single pass.

**6. `failureFunction` writes vision-failed status even when failure was in the image-edit step** — `src/lib/workflows/replace-element-workflow.ts:222-272`

If image-edit fails (not vision), the element card still shows \"Vision failed: <image-edit error>\" because the failureFunction writes `visionStatus: 'failed'` for every workflow failure. Misleading. A separate `replaceStatus` field — or branching the failureFunction on which step failed — would help. Also, the catch around `updateVisionStatus` (line 267) only `console.error`s; a DB write failure here means the element is stranded in `analyzing` with no DB-level evidence of failure (mirrors the Critical above).

### Nits

- **Storage-path validation** at `src/functions/sequence-elements.ts:274` (also pre-existing on line 145) uses `startsWith` only, accepting `elements/<myTeamId>/../<otherTeamId>/foo`. R2 stores literal keys so the practical blast radius is limited today, but tightening to an exact `buildR2Key` recompute or `..`/`//` rejection would close the namespace boundary.
- `src/hooks/use-sequence-elements.ts:249` invalidates `['frames']` (very broad). Worth checking whether the recast hooks invalidate more narrowly.
- `src/components/element/replace-element-confirm-dialog.tsx:39-58` nests `<span className=\"block\">` + `<strong>` inside `AlertDialogDescription` (which renders a `<p>`). Technically valid HTML, but cleaner to render the secondary lines outside the description element.
- `replace-element-workflow.ts` uses `console.log`/`console.error` rather than the project-standard `logError` for Sentry tracking. Peer workflows mostly do the same, so this is consistent — flag only if your error-tracking standard tightens.
- `buildEditPrompt` (`replace-element-workflow.ts:40-58`) is the one pure helper worth a small unit test (3 branches: previousDescription present/absent, newDescription present/absent, token interpolation). Skip everything else — workflow-shell tests aren't the project's convention.

Issue #10, opened 2025-08-28
[Backend] Characters: Implement character library with LoRA support


tombeckenham:
## Description
Build the character management system that handles LoRA models for consistent character appearance across frames.

## Acceptance Criteria
- [ ] Character CRUD operations
- [ ] LoRA file upload to storage
- [ ] Character metadata management
- [ ] Character-frame associations
- [ ] Weight adjustment per usage
- [ ] Character gallery per team
- [ ] Training status tracking
- [ ] Character usage analytics
- [ ] File validation and security checks
- [ ] Storage quota enforcement

## Technical Implementation Notes
- Store LoRA files in Supabase Storage
- Validate file formats and sizes
- Generate thumbnails for characters
- Track usage for optimization
- Implement virus scanning for uploads

## API Endpoints
- POST /api/v1/characters - Create character
- GET /api/v1/characters - List team characters
- GET /api/v1/characters/:id - Get character details
- PUT /api/v1/characters/:id - Update character
- DELETE /api/v1/characters/:id - Delete character
- POST /api/v1/characters/:id/upload - Upload LoRA file
- POST /api/v1/characters/:id/train - Start training job
- GET /api/v1/characters/:id/usage - Get usage stats

## Database Schema
```sql
characters (
  id uuid primary key,
  team_id uuid references teams,
  name varchar,
  description text,
  lora_url text,
  thumbnail_url text,
  training_images jsonb,
  config jsonb,
  status varchar, -- draft, training, ready, failed
  file_size_bytes bigint,
  usage_count integer,
  created_at timestamp,
  updated_at timestamp
)

character_training_jobs (
  id uuid primary key,
  character_id uuid references characters,
  job_id uuid references jobs,
  training_params jsonb,
  status varchar,
  error text,
  created_at timestamp,
  completed_at timestamp
)
```

## Dependencies
- Team management (#5)
- QStash queue (#2)
- Storage setup (part of #1)

## Estimated Complexity
**Story Points**: 13
**Time Estimate**: 3 days

PR #720, opened 2026-05-17
Styles foundation: schema, popularity tracking, recommendation badges #716

### LINKED ISSUE #716: Styles foundation: schema, popularity tracking, recommendation badges
## Context

Foundation work for the 100-styles expansion. Today the `styles` table (`src/lib/db/schema/libraries.ts`) has a `usageCount` column that is never incremented, no notion of sample videos, and no way for a style to recommend an image model, motion model, or aspect ratio. This issue puts all the plumbing in place so the content + discovery work can land on top.

Plan: `/Users/tom/.claude/plans/i-want-to-expand-sunny-pancake.md`

### Schema additions on `styles`
- `sampleVideos` (JSON): `Array<{ url, kind: 'canonical' | 'category' | 'bespoke', label, durationSeconds, order }>`
- `recommendedImageModel` (nullable text, validated against `src/lib/ai/models.ts`)
- `recommendedVideoModel` (nullable text, validated against `src/lib/ai/models.ts`)
- `defaultAspectRatio` (nullable text, validated against `src/lib/constants/aspect-ratios.ts`)
- `useCases` (JSON string[], default `[]`) plus a controlled vocabulary file next to `style-templates.ts` (initial values: `talking-head`, `product`, `lifestyle`, `b-roll`, `animatic`, `kids`, `tutorial`, `pitch-deck`, `social-vertical`)
- Generate migration with `bun db:generate` — do not hand-write SQL
- Update `StyleConfig` Zod schema and the `Omit<…>` in `src/lib/style/style-templates.ts` so existing templates still type-check (all new fields optional)

### Popularity wiring
- Add `incrementUsage(styleId)` to `src/lib/db/scoped/styles.ts` — atomic `UPDATE styles SET usageCount = usageCount + 1`. Global, no team scope check.
- Add `incrementStyleUsageFn` server function in `src/functions/styles.ts`
- Fire from the server handler that sets `sequence.styleId` (grep `styleId:` under `src/routes/api/sequences/`); also `posthog.capture('style_selected', { styleId, sequenceId, teamId })` for future time-windowed popularity
- Extend `getStyles` scoped query with optional `orderBy: 'popular' | 'sortOrder'` (default stays `sortOrder`)

### Recommendation badges (suggest, never auto-change)
- `MotionModelSelector` (`src/components/model/motion-model-selector.tsx`): accept `recommendedVideoModel`, render badge on matching option. Tooltip: \"Recommended for {style.name}\"
- `ImageModelSelector` (`src/components/model/image-model-selector.tsx`): same treatment for `recommendedImageModel`
- `AspectRatioPills` in `GenerationSettings`: badge driven by `style.defaultAspectRatio`
- Plumb the style fields down from `scenes-view.tsx` (already calls `useStyle(sequence.styleId)`)
- Do not auto-select or override the user's current choice

## Verification

- `bun db:generate && bun db:migrate` succeed locally
- `bun tsgo --noEmit` clean
- Create a sequence, pick a style → row in DB shows `usageCount` incremented and PostHog event lands
- Pick a style with `recommendedVideoModel` set → \"Recommended\" badge appears; switching models works and doesn't snap back
- Minimal new tests: one for `incrementUsage`, one to confirm `getStyles({ orderBy: 'popular' })` returns ordered correctly

## Out of scope (separate issues)
- New style templates (content)
- Thumbnails / sample videos (asset pipeline)
- Discovery UX changes (popular section, hover video, use-case pills)

tombeckenham:
## Related Issue
Closes #716

## Summary
Implemented styles foundation: schema additions + migration, popularity wiring (incrementUsage + posthog), recommendation badges across image/motion/aspect-ratio selectors, tests; pushed branch for PR.

## Notes

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,8 +1,6 @@
 # CLAUDE.md
 
-This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
-
-AI-powered video sequence platform built with TanStack Start, optimized for edge deployment.
+AI-powered video sequence platform built with TanStack Start, deployed to Cloudflare Workers.
 
 ## Commands
 
@@ -14,19 +12,19 @@ bun db:studio                      # Drizzle Studio for local DB
 
 # Quality
 bun lint                           # oxlint (type-aware)
-bun lint:fix                       # oxlint with autofix
-bun format                         # oxfmt write
-bun format:check                   # oxfmt check (CI)
+bun lint:fix
+bun format                         # oxfmt
+bun format:check
 bun typecheck                      # tsgo --noEmit (NOT `tsc`)
 bun dead-code                      # knip (unused exports)
 
 # Tests
 bun test                           # unit (bun:test)
 bun test path/to/file.test.ts      # single file
-bun test --watch                   # watch mode
-bun test --coverage
+bun test:watch
+bun test:coverage
 bun test:e2e                       # Playwright
-bun test:e2e:ui                    # Playwright UI
+bun test:e2e:ui
 bun test:e2e:setup                 # rebuild test.db + seed
 
 # DB
@@ -39,7 +37,7 @@ bun run build                      # Vite production build (note: NOT `bun build
 bun cf:deploy:prd                  # Cloudflare Workers production deploy
 ```
 
-`bun dev` runs everything in parallel — there is no separate `qstash:dev` terminal anymore (Docker is required so QStash can start).
+`bun dev` runs everything in parallel via Docker (QStash) — there is no separate `qstash:dev` terminal anymore.
 
 ## Project Structure
 
@@ -61,27 +59,18 @@ scripts/            # CLI tooling and setup
 drizzle/migrations/ # Generated SQL (do NOT hand-edit)
 ```
 
-## Architecture Overview
+## Architecture
 
-**Tech Stack:**
+**Stack:** Bun · TanStack Start + Router + Vite · Turso (libSQL/SQLite) + Drizzle · QStash (durable workflows) · Cloudflare R2 · Better Auth · Tailwind v4 + shadcn/ui · `bun:test`.
 
-- **Runtime**: Bun (not Node.js)
-- **Framework**: TanStack Start + TanStack Router + Vite
-- **Database**: Turso (libSQL/SQLite) + Drizzle ORM
-- **Workflows**: QStash (durable execution for AI tasks)
-- **Storage**: Cloudflare R2 (S3-compatible)
-- **Auth**: Better Auth
-- **Styling**: Tailwind v4 + shadcn/ui
-- **Testing**: Bun test
+**Core rules:**
 
-**Core Principles:**
+- Database access ONLY in server handlers (never in components).
+- Anonymous-first → upgrade to save work.
+- Team-based resources (sequences, styles, characters).
+- Script-driven generation for consistency.
 
-- Database access ONLY in server handlers (never in components)
-- Anonymous-first → upgrade to save work
-- Team-based resources (sequences, styles, characters)
-- Script-driven generation for consistency
-
-**Data Model:**
+**Data model:**
 
 ```
 teams
@@ -91,8 +80,6 @@ teams
   └── libraries (styles, characters, vfx, audio)
 ```
 
----
-
 ## Setup
 
 ```bash
@@ -101,55 +88,40 @@ bun setup                          # Auto-configure local dev (SQLite + QStash)
 bun db:setup                       # Migrate + seed database
 ```
 
-**Daily workflow:** `bun dev` — runs DB migrate/seed, Vite, QStash (Docker), and Stripe listener in parallel.
-
-**Branch + commit conventions:** Branches must be named `<issue-number>-feature-name` (e.g. `393-improve-readme`). Lefthook extracts the issue number and tags commits with `#<issue>` automatically. See `CONTRIBUTING.md`.
-
-**Before commit:** Lefthook auto-checks quality.
+**Branch + commit conventions:** Branches must be named `<issue-number>-feature-name` (e.g. `393-improve-readme`). Lefthook extracts the issue number and tags commits with `#<issue>` automatically. See `CONTRIBUTING.md`. Lefthook also runs quality checks pre-commit.
 
 ---
 
 ## Server Handler Pattern
 
-All API routes use TanStack Start server handlers:
+All API routes use TanStack Start server handlers. Standard shape:
 
 ```typescript
 // src/routes/api/example/$id.ts
-import { createFileRoute } from '@tanstack/react-router';
-import { json } from '@tanstack/react-start';
-import { requireUser } from '@/lib/auth/action-utils';
-import { handleApiError } from '@/lib/errors';
-
 export const Route = createFileRoute('/api/example/$id')({
   server: {
     handlers: {
       POST: async ({ params, request }) => {
         try {
-          // 1. Validate input
           const input = schema.parse(await request.json());
-
-          // 2. Check auth/team permissions
           const user = await requireUser();
 
-          // 3. Execute business logic (DB operations ONLY here)
-          const record = await db.insert(table).values({
-            ...input,
-            teamId: user.teamId,
-          });
+          const record = await db
+            .insert(table)
+            .values({ ...input, teamId: user.teamId });
 
-          // 4. Trigger workflows for async AI tasks
+          // Trigger workflows via QStash (see Workflow Pattern below)
           const { messageId } = await qstash.publishJSON({
             url: `${getQStashWebhookUrl()}/workflows/image`,
             body: { userId: user.id, teamId: user.teamId, ...input },
           });
 
-          // 5. Return standardized response
           return json({ id: record.id, workflowRunId: messageId });
         } catch (error) {
-          const handledError = handleApiError(error);
+          const handled = handleApiError(error);
           return json(
-            { success: false, error: handledError.toJSON() },
-            { status: handledError.statusCode }
+            { success: false, error: handled.toJSON() },
+            { status: handled.statusCode }
           );
         }
       },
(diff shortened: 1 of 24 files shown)


tombeckenham (commented):
## Automated review — round 1

Three specialized reviewers ran in parallel over the diff. Verdict: **needs changes** — several correctness issues on the popularity-counter path and a silently-dropped-fields bug in `createStyleFn`. The schema/migration work and the badge plumbing are otherwise solid (no table rebuild, ADD COLUMN only — avoids the D1/Turso CASCADE trap from CLAUDE.md).

### 🔴 Critical

**1. `createStyleFn` silently drops the new fields** — `src/functions/styles.ts:75-84`
`createStyleSchema` accepts `sampleVideos`, `useCases`, `recommendedImageModel`, `recommendedVideoModel`, `defaultAspectRatio` (all optional via `createInsertSchema`), but the handler builds the insert payload from a hand-picked subset (`name`, `description`, `config`, `category`, `tags`, `isPublic`, `previewUrl`). Any caller setting recommendation fields gets a successful response with the values silently discarded. Spread `data` or whitelist the new fields explicitly. (`updateStyleFn` at line 105 is fine — it spreads `...updateData`.)

**2. `incrementStyleUsageFn` is an open popularity-counter abuse vector** — `src/functions/styles.ts:136-142`
Endpoint accepts any `styleId` from any authenticated team member, increments `usageCount` globally, returns `{ success: true }`. No existence check, no visibility check (team-owned or public), no rate limit. Any user can spam this to inflate any style's popularity ranking — including styles their team can't see — and now `list({ orderBy: 'popular' })` drives recommendation surfaces. Either gate behind `getById` + visibility check (`teamId === context.teamId || isPublic`), or drop the standalone endpoint and only increment server-side from `createSequenceFn`/`updateSequenceFn` (the only legitimate triggers today).

**3. Popularity counter and PostHog event fire N times per sequence creation** — `src/functions/sequences.ts:170-176`
Both `incrementUsage` and `trackStyleSelected` are inside `Promise.all(analysisModels.map(async ...))`. When the user picks 3 analysis models, one click bumps `usageCount` by 3 and emits 3 `style_selected` events. This skews both the recommendation ranking and analytics. Hoist both calls above the `.map(...)` so they fire once per `createSequenceFn` invocation.

**4. Awaited side-effects abort sequence creation/update on transient failure** — `src/functions/sequences.ts:170`, `:269`
`await context.scopedDb.styles.incrementUsage(...)` in `createSequenceFn` runs after the sequence row is inserted. A DB blip throwing here rejects the `Promise.all` lambda, abandons peer sequences mid-flight (some persisted but without workflows triggered, some workflows triggered without `promoteTempElements` running), and the user sees a 500 even though several sequences exist. Same shape in `updateSequenceFn:269` — DB row is already updated, but a popularity-write failure rejects the response and lets React Query think the update failed. Wrap in `.catch(logError)` or `void`-ignore — popularity must never block the critical path. (`trackStyleSelected` has the same blast-radius problem at sequences.ts:49 — `posthog.capture` is unwrapped.)

### 🟡 Important

**5. `incrementUsage` silently no-ops on unknown styleId** — `src/lib/db/scoped/styles.ts:89-94`
`db.update().where(eq(styles.id, styleId))` matching zero rows is a successful query with no signal. Typos, deleted styles, or cross-team styleId leaks produce zero updates with no error. Check `rowsAffected` (or use `.returning()`) and at minimum log when zero rows match.

**6. `incrementStyleUsageFn` returns `{ success: true }` regardless of effect** — `src/functions/styles.ts:140-142`
Combined with #5, the caller can never distinguish "incremented" from "styleId didn't exist". This also violates CLAUDE.md's "throw errors instead of returning success boolean" rule (root CLAUDE.md, *Global Rules*).

**7. Recommendation badge silently vanishes when key doesn't match** — `src/components/model/image-model-selector.tsx:32-37`, `motion-model-selector.tsx:56-61`, `aspect-ratio-pills.tsx:52-55`
If a style's `recommendedImageModel` / `recommendedVideoModel` / `defaultAspectRatio` doesn't match any rendered item (deleted model id, motion model filtered out by aspect-ratio compat at motion-model-selector.tsx:46-48, ratio not in `ASPECT_RATIOS`), no badge renders and no warning fires. A style admin sets a recommendation, it sticks in the DB, never appears. Motion is especially load-bearing: aspect-ratio compatibility can hide a valid recommendation — that case probably wants surfacing as "recommended for this style, not compatible with current ratio" rather than no badge at all.

**8. Migration adds nullable columns but `$type` lies on existing rows** — `drizzle/migrations/20260517011119_…/migration.sql`
`sampleVideos` and `useCases` are typed as `$type<StyleSampleVideo[]>` / `$type<string[]>` with `$defaultFn(() => [])`, but the SQL columns have no `DEFAULT '[]'`. Existing pre-migration rows are NULL while the inferred TS type says non-null. No current reader assumes non-null (only writes today), so latent — but the contract is wrong. Either backfill (`UPDATE styles SET sample_videos='[]' WHERE sample_videos IS NULL` in the migration) or change the types to `… | null`.

**9. Test inline-redefines methods instead of exercising production code** — `src/lib/db/scoped/styles.test.ts:34-76`
`makeStylesMethods` is an inline copy of `createStylesMethods` rather than an import. The justification on lines 30-33 (mock leakage from `scoped.test.ts`) is incorrect — Bun runs each test file in its own process, so the global `mock.module` does not actually leak here. Net effect: a typo or `eq` swap in production `src/lib/db/scoped/styles.ts:49-96` will currently pass the test suite. Drop the inline copy, import `createStylesMethods` directly.

**10. New endpoint and new write paths untested**
- `incrementStyleUsageFn` has no test (authz model deserves at least one coverage point).
- `updateSequenceFn`'s new `previousStyleId !== updateData.styleId` guard has no test — silent double-counting or missed increments would surface only in production.
- No round-trip test for the new schema fields (`sampleVideos`, `recommendedImageModel`, etc.).

### ⚪ Nits

- `src/lib/db/scoped.test.ts:128-135` — `createStylesMethods` mock omits `incrementUsage`. No caller hits it today, so latent — but add `incrementUsage: mock()` while the context is fresh, otherwise the next test author writing `sequences.test.ts` will hit `TypeError: not a function` and have to dig.
- `src/components/model/image-model-selector.tsx:62` — `isValidTextToImageModel(m.id)` filter is redundant since `m.id` already comes from `Object.entries(IMAGE_MODELS)`.
- `src/lib/db/scoped/styles.test.ts:121` — test labelled "atomically" but issues sequential awaited calls. Either rename to "bumps usageCount by 1 on each call" or add a `Promise.all([...10 increments])` test that asserts final count = 10.
- `src/lib/db/scoped/styles.test.ts` — `list({ orderBy: 'popular' })` doesn't cover the `isPublic: true` visibility branch.

tombeckenham (commented):
## Automated review — round 2

Round-1 findings: most addressed (see table). One round-1 finding closed at the endpoint level was reopened through a different surface by the same round-2 commit. That is the blocker.

### 🔴 Critical

**1. createStyleFn / updateStyleFn allow client to set usageCount, isTemplate, version, sortOrder** — src/functions/styles.ts:72-78, :93-108 + src/lib/schemas/style.schemas.ts:17-28

Round 1 dropped the standalone incrementStyleUsageFn to close the popularity-inflation hole. Round 2 fixed the silently-dropped-fields bug (good) by spreading the schema payload after stripping teamId/createdBy. But createInsertSchema(styles, ...) and createUpdateSchema(styles, ...) include every column on the table — usageCount, isTemplate, version, sortOrder, id, createdAt, updatedAt. scopedDb.styles.update at src/lib/db/scoped/styles.ts:71-81 does an unfiltered .set(data).

Net effect: a team admin POSTs { styleId, usageCount: 999999 } to updateStyleFn and pumps their style to the top of list({ orderBy: popular }). Same hole round 1 closed, surfaced through a different door.

Fix: add .omit({ id, teamId, createdBy, createdAt, updatedAt, usageCount, version, isTemplate }) (or equivalent) to both schemas. Don't rely on per-handler destructuring or TypeScript types — Zod is the only runtime gate.

### 🟡 Important

**2. Test inline-copy of incrementUsage has a different return type than production** — src/lib/db/scoped/styles.test.ts:71-78 vs src/lib/db/scoped/styles.ts:89-98

The Bun mock.module rationale (lines 30-35, oven-sh/bun#7823) is legitimate and mirrors precedent at src/lib/db/scoped/talent.test.ts:11-12 — keep the workaround. But the inlined incrementUsage returns Promise<number> while production returns Promise<void> and instead logs console.warn on zero rows. The "returns zero affected rows" assertion at line 142-146 tests behavior that doesn't exist in production. A future engineer who removes the warn won't get a test failure. Pick one: (a) change production to Promise<number> and let bumpStylePopularity log with full context, or (b) keep production as-is and spy on console.warn in the test.

**3. Image/aspect recommendation badge still silently vanishes for users** — src/components/model/image-model-selector.tsx:91-100, src/components/settings/aspect-ratio-pills.tsx:39-49

Round-1 finding #7 only got fixed for developers (console.warn) and for motion (visible "incompatible with current aspect ratio" hint at motion-model-selector.tsx:111-118). For image-model and aspect-ratio pills, a style admin who sets recommendedImageModel: flux_pro_deprecated_v0 sees no badge, no warning in the UI, and no signal anywhere except a client-side console.warn that nobody will ever see in production. Either mirror motion's inline hint, or constrain recommendedImageModel / defaultAspectRatio to a known-key enum at the schema layer so unmatched recommendations can't be persisted.

**4. bumpStylePopularity and its invariants are entirely untested** — src/functions/sequences.ts:42-72, :170-178, :283-294

The one-bump-per-click invariant at sequences.ts:243-244 is held together by a comment. A future refactor that moves the call inside the Promise.all(...map) re-introduces Nx inflation silently. The previousStyleId !== updateData.styleId guard at sequences.ts:283-286 is the only defense against PATCH-loops inflating the counter. Both deserve dedicated tests. Suggested cases:
- createSequenceFn with three analysis models → incrementUsage called exactly once.
- updateSequenceFn with styleId === previousStyleId → not called; with a new styleId → called once.
- createSequenceFn resolves even when mocked incrementUsage rejects (proves .catch actually swallows).

**5. PostHog try/catch is dead code** — src/functions/sequences.ts:58-72

posthog-node capture() returns synchronously and already swallows internal errors via its own .catch((err) => console.error(err)) inside addPendingPromise. The OpenStory try/catch will only ever catch a getPostHogClient() constructor throw — the SDK's actual network/queue failures will surface only as untagged console.error lines from posthog-node. Either drop the wrapper (it's noise that implies coverage that doesn't exist) or wire posthog.on(error, handler) once at client creation so capture failures get an OpenStory-tagged log.

**6. bumpStylePopularity log strips the most useful debugging fields** — src/functions/sequences.ts:52-57

userId, teamId, and sequenceIds.length are all in scope but not included. If popularity stops incrementing in prod (the reason this PR exists), the only signal is "[styles] incrementUsage failed { styleId, err }" with no team/user correlation. Suggest console.error (not warn) with { styleId, teamId, userId, sequenceCount, err }. Bonus: production incrementUsage ends with the zero-row warn but the caller can't enrich it — change the return type so this becomes possible (see #2).

### Nits

- src/functions/sequences.ts:286 — the "&& sequence.styleId" guard is dead per the notNull typing on sequences.styleId. Either drop, or comment why the runtime guard is needed (FK has onDelete set null — pre-existing).
- useEffect warns fire on every render where the recommendation doesn't match. If styleName is a fresh string identity on every parent render this becomes a log flood. Move check to a useEffect keyed on the deps that actually matter (memoized) or fire once via a ref.
- bumpStylePopularity .catch(...) returns a Promise<void> that is discarded — void-prefix the expression so a future maintainer who adds a re-throw in the catch handler doesn't silently create an unhandled rejection.
- The migration comment says "match the inferred non-null TS types" but sampleVideos/useCases are typed as nullable in InferSelectModel because the columns are not .notNull(). The backfill is correct; just rephrase the comment.

### Round-1 status

- #1 createStyleFn drops new fields: Fixed (but see Critical #1 — different hole)
- #2 incrementStyleUsageFn endpoint: Endpoint removed (but see Critical #1)
- #3 N-times popularity bump per sequence creation: Hoisted to once-per-click
- #4 Awaited side-effects abort sequence creation: Fire-and-forget via .catch()
- #5 incrementUsage silent zero-row no-op: Warn added (but see Important #2)
- #6 Returns success boolean: Endpoint dropped
- #7 Recommendation badge silently vanishes: Partial — motion fixed, image/aspect log-only (Important #3)
- #8 Migration missing JSON backfill: Backfill added
- #9 Test inline-redefines production methods: Justified, but signature drift (Important #2)
- #10 Missing tests: DB-layer tests added; bumpStylePopularity invariants still uncovered (Important #4)

PR #732, opened 2026-05-19
Scene prompt editor: @-mention autocomplete for elements, cast, locations #695

### LINKED ISSUE #695: Scene prompt editor: @-mention autocomplete for elements, cast, locations
Follow-up to #683 (which lands the server-side auto-link on save).

## Problem
After #683, users can type a canonical tag (element token, characterId/consistencyTag, locationId/consistencyTag) in the scene image- or motion-prompt and the system links the reference on save. But the user has no in-editor way to discover what tags exist — they have to remember or look up the library in another tab.

## Proposed solution
Add `@`-mention autocomplete inside the image-prompt and motion-prompt textareas on `src/components/scenes/scene-script-prompts.tsx`:

- Typing `@` opens a popover anchored to the caret.
- Popover lists the sequence's elements, cast, and locations (sectioned) with their canonical tag and a thumbnail where available.
- Filter as the user types.
- Selecting an entry inserts the canonical tag at the caret (the same string that the #683 server-side parser already recognises).
- Keyboard: ↑/↓ to navigate, Enter/Tab to insert, Esc to dismiss.

## Notes
- The parser/match logic from #683 (`src/lib/workflows/extract-continuity-from-prompt.ts`) already accepts the canonical forms this UI will emit — no server-side changes needed.
- Reuse the existing sequence-level lists (`scopedDb.characters.list`, `sequenceElements.list`, `sequenceLocations.list`).

tombeckenham:
## Related Issue
Closes #695

## Summary
Implemented @-mention autocomplete in scene image/motion prompts, with sectioned popover, keyboard nav, unit tests; pushed branch — awaiting CI + review.

## Notes


tombeckenham (commented):
## Automated review — round 1

Scope: 6 files, +679/-7. New module under `src/components/scenes/prompt-mention/` (3 source + 2 tests) and a wiring change in `scene-script-prompts.tsx`. Pure-TS logic is tidy and has unit coverage; the React shell composes Radix Popover + Textarea cleanly. **No critical issues** — round-trip with `extract-continuity-from-prompt.ts` is correct (element tokens are case-insensitive in `matchElementsToScene`, character/location tags use the same slug shape the workflow recognises).

### Important

**1. Use `@unpic/react` instead of raw ``**
`src/components/scenes/prompt-mention/prompt-mention-textarea.tsx:253-257`
Project convention is `@unpic/react` for all images; the codebase already uses it for similarly sized thumbnails (`scene-thumbnail.tsx`, `style-selector.tsx`). 32×32 popover thumbnails would benefit from the configured CDN optimisation and avoid CLS.

**2. Incomplete WAI-ARIA combobox pattern**
`src/components/scenes/prompt-mention/prompt-mention-textarea.tsx:170-177`, `:243-281`
The textarea sets `aria-autocomplete="list"` + `aria-expanded` + `aria-controls`, but is missing `role="combobox"` on the input, `aria-activedescendant` pointing at the active item, and `role="listbox"` / `role="option"` / `aria-selected` on the list. The current `<button aria-current="true">` is not the listbox-option pattern; screen-reader users get no announcement on ArrowUp/Down. CLAUDE.md requires "Full keyboard support per WAI-ARIA APG".

**3. Hard-coded popover id collides under reuse**
`src/components/scenes/prompt-mention/prompt-mention-textarea.tsx:176, 180`
`id="mention-popover"` is safe today because Radix `Tabs` unmounts inactive `TabsContent`, but the component is exported as reusable. Two instances mounted simultaneously (e.g. side-by-side preview, future modal) would produce duplicate DOM ids and broken `aria-controls`. Use `useId()`.

### Nits

- `insertMention` always appends a trailing space; when the caret is followed by `" later"` the result is two consecutive spaces (`mention-trigger.test.ts:64` bakes this in as `"see jack-denim-jacket  later"`). Consider trimming a leading space from `after` if the next char is whitespace. (`src/components/scenes/prompt-mention/mention-trigger.ts:51-55`)
- `useImperativeHandle` (`prompt-mention-textarea.tsx:51-56`) is dead — no caller passes a `ref`, and the `throw new Error(...)` path inside it can't fire in practice. Either drop the `ref` prop or just forward `textareaRef.current`. CLAUDE.md: "Don't add error handling for scenarios that can't happen."
- No round-trip integration test asserting that `insertMention(...)` output is parsed correctly by `extractContinuityFromPrompt`. The whole feature's value rests on that contract; one fixture-based test would lock it in.
- `mention-trigger.ts:27-29`: `if (ch === undefined) return null;` and the `prev && ...` guard inside the `@` branch are dead (loop bounds guarantee `ch` is defined and `text[i-1]` is a string when `i !== 0`).

Format check, types, and tests should pass under lefthook; nothing here blocks merge.

PR #745, opened 2026-05-26
fix(merge): hold status until mux + replace compare dialog with history sheet (#741)

### LINKED ISSUE #741: Fix merged-video "wrong final" bug + replace variant compare dialog with a history view
## Context

Two related issues with the merged-video variant flow surfaced while tracing why the wrong version sometimes shows after a merge:

1. **A real bug** in the merge → audio-mux handoff that briefly (or permanently, after a promote) shows the bare merged video without music.
2. **The compare dialog is doing a job that a history view would do better.** Only one alternate is ever surfaced; the "live vs alternate" framing is confusing; the rest of the kept variants are invisible.

Both touch the same code (`merge-video-workflow.ts`, `merge-audio-video-workflow.ts`, `sequence-variants.ts`, `sequence-variant-compare-dialog.tsx`), so bundling them.

---

## Bug 1 — `mergedVideoUrl` is set to the bare merge before mux runs

`src/lib/workflows/merge-video-workflow.ts:285-302` (multi-video path) and `:153-167` (single-video path) do:

```ts
await seq.updateMergedVideoFields({
  mergedVideoUrl: storageResult.url,  // bare merge, no music
  mergedVideoStatus: 'completed',
  ...
});
await getGenerationChannel(...).emit('generation.merge:progress', {
  step: 'video', status: 'completed', mergedVideoUrl: storageResult.url,
});
await chainAudioMux(...);  // only NOW does merge-audio-video run
```

`merge-audio-video-workflow.ts:48-55` then flips status back to `'merging'`, muxes, and finally writes the muxed URL. The theatre view reads `sequence.mergedVideoStatus === 'completed' && mergedVideoUrl` (`theatre-view.tsx:91`) so it has no way to know "completed (no music)" is actually a transient state — the user sees a music-less final, sometimes for a while if QStash is slow or the mux step errors.

### Fix

When music is `'completed'` for the sequence and `chainAudioMux` will run, **don't write `mergedVideoUrl` / `mergedVideoStatus = 'completed'` in the merge-video step.** Keep `mergedVideoStatus = 'merging'` and let `merge-audio-video-workflow` be the one that flips to completed with the muxed URL. Same change on both the single-video (line 153) and multi-video (line 285) paths.

The bare URL still gets stored on the variant row (so the variant table is complete), but `sequences.merged*` only updates once for the lifecycle. Pull the music-completed check out of `chainAudioMux` into the caller so we can branch on it.

## Bug 2 — Promoting a divergent merged video drops music

Same root cause. `sequence_video_variants` rows hold the **bare** merged URL (because only `merge-video-workflow` writes variants). `promoteVideoVariant` (`src/lib/db/scoped/sequence-variants.ts:306-359`) copies `variant.url` straight onto `sequences.mergedVideoUrl` and never re-mux audio.

### Fix

After `promoteVideoVariant` commits, if the sequence has `musicStatus === 'completed'` and a music primary variant, kick off `merge-audio-video-workflow` with the promoted video variant id + current music primary variant id. Until that completes, `mergedVideoStatus` should be `'merging'` (set in the same batch that does the promote).

Add a test in `sequence-variants.test.ts` asserting `promoteVideoVariant` does not leave the sequence in `completed` state when music is present.

---

### Why drop compare

`src/components/sequence/sequence-variant-compare-dialog.tsx` only surfaces **one** alternate — `divergentVideoVariants?.[0]`, the oldest pending one (`listDivergentVideos` orders by `divergedAt` asc). All other kept variants are invisible. The "Live (current inputs) vs Alternate (older inputs)" framing inverts the user's mental model — they're thinking "which one is the final I want," not "which one matches today's inputs." For video there's no real diff content either; the music side has a small diff table but video is just two players.

### Replacement: variant history sheet

A right-aligned sheet (`<Sheet>` from shadcn) titled "Merged video history" with one row per variant for `(sequenceId, workflow='merge-video')`:

- **Current** badge on the primary row (the one matching `sequences.mergedVideoUrl`).
- Per row: thumbnail/preview button, `generatedAt`, `inputHash` short, an "Inputs changed" chip when `divergedAt IS NOT NULL`, and overflow menu with **Preview**, **Promote**, **Discard / Undo**.
- Ordering: `generatedAt` DESC. (`listDivergentVideos` currently orders by `divergedAt` ASC for the banner — keep that endpoint, add a new `listVideoVariantsBySequence` that returns everything including the primary, ordered newest first.)
- Promote action keeps the existing destructive-confirm pattern (sonner toast + Undo).
- Inline expand for side-by-side preview vs. current — replaces the compare dialog's job, but as an in-row action, not the entry point.

### Banner stays, points at the history

`DivergentAlternateBanner` keeps its current role: surfaces when a new divergent alternate appears. Its CTA changes from "Compare" → "View history" and opens the sheet. The "you have an alternate you haven't decided on" prompt is still useful — we're just moving where the decision happens.

### Music

Same treatment for music variants — replace the music side of the compare dialog with a parallel history view triggered from `_protected/sequences/$id/music.tsx`. Music variants do have meaningful diff fields (prompt/tags/model/duration); surface them as per-row chips in the history rather than a separate dialog panel.

### Code touchpoints

- **Add**: `src/components/sequence/sequence-variant-history-sheet.tsx`, `useSequenceVideoVariantHistory` / `useSequenceMusicVariantHistory` hooks in `src/hooks/use-sequence-variants.ts`, `listVideoVariantsBySequence` / `listMusicVariantsBySequence` reader endpoints + server functions.
- **Modify**: `DivergentAlternateBanner` CTA wiring, `theatre.tsx` and `music.tsx` routes to mount the sheet instead of the dialog, query-cache-updater `generation.stale:detected` handler (already invalidates `sequence-divergent-video` — extend to invalidate the new full-list key).
- **Remove**: `src/components/sequence/sequence-variant-compare-dialog.tsx` and its stories.

---

## Out of scope

- Diffing the *inputs* themselves (which frame changed, which sheet diverged). Today we only have `inputHash` strings; surfacing semantic input diffs is its own feature.
- Versioning earlier renders that were overwritten in-place (`upsertVideoPrimary` clobbers — only divergence keeps history). Retaining all renders would need a schema change.

## References

- `src/lib/workflows/merge-video-workflow.ts:153-167, 285-302` — bare-URL write before mux
- `src/lib/workflows/merge-audio-video-workflow.ts:48-141` — mux writes
- `src/lib/db/scoped/sequence-variants.ts:210-358` — variant write/promote/discard
- `src/components/sequence/sequence-variant-compare-dialog.tsx` — to be removed
- `src/components/theatre/theatre-view.tsx:91` — current consumer of `sequences.mergedVideoUrl`

tombeckenham:
## Summary

Two related fixes for the merged-video flow plus a UI refresh.

- **Bug 1 — soundless final between merge and mux.** `merge-video-workflow` was setting `mergedVideoUrl` + `mergedVideoStatus='completed'` *before* `merge-audio-video` ran, so the theatre view briefly (sometimes permanently, if mux failed) showed the bare merge. The mux-music check now runs before the sequence update; when mux will fire we hold `mergedVideoStatus='merging'` across the QStash hop and let `merge-audio-video` flip status to completed with the muxed URL. Applies to both single-video and multi-video paths.
- **Bug 2 — promote dropped music.** `promoteVideoVariant` copied the bare variant URL straight onto `sequences.mergedVideoUrl`, even when music was present. Now: if a music primary exists, promote sets `mergedVideoStatus='merging'` (keeping the previous muxed URL in place), then `promoteSequenceVideoVariantFn` fires `merge-audio-video` so the bare merge is muxed against the current music before publishing.
- **Compare dialog → history sheet.** The old dialog only surfaced the oldest pending divergent alternate. The new right-aligned `Sheet` lists every render for a sequence (primary + divergent + discarded) with Current / Inputs-changed / Discarded chips and per-row Promote/Discard/Restore. Same shell for video and music. `DivergentAlternateBanner` CTA switches from "Compare" → "View history" for sequence-level use.
- **Auto-merges replace the primary.** Adding frames to an in-progress sequence shouldn't surface as "inputs changed" — that fork-vs-promote decision matters when the user manually re-triggers a merge mid-flight, not when motion-batch auto-merges because frames just finished. `motion-batch-workflow` and `motion-workflow`'s `check-merge-trigger` now set `replacePrimary: true`, which bypasses divergence routing in `writeVideoVariant`. Manual triggers (`mergeVideoAndMusicFn`, `triggerMergeVideoFn`) keep the existing divergence behavior.

## Test plan

- [ ] Theatre view: kick off a multi-frame merge with music ready → status stays `merging` until the muxed URL lands; bare merge never appears as the final.
- [ ] Generate frames in two passes (e.g. 4, then add 2) → second auto-merge replaces the primary, no "inputs changed" alternate created.
- [ ] Manually re-trigger merge after changing an upstream input → divergent alternate still surfaces and the history sheet lets you promote it.
- [ ] Promote a divergent alternate when music is present → mux fires, theatre player shows the muxed final once status flips back to completed.
- [ ] History sheet: opens from "View history" CTA, lists all rows newest-first, Current badge on the live row, per-row preview, promote/discard/restore work.
- [ ] `bun test src/lib/db/scoped/sequence-variants.test.ts` green (new regression tests for `replacePrimary` + `pendingAudioMux`).


tombeckenham:
This has been replaced by the new mediabunny workflow. Closing

PR #851, opened 2026-06-07
Remove Langfuse — LLM analytics via TanStack AI OTel → PostHog

### LINKED ISSUE #849: Remove langfuse
We don't use langfuse anymore, only posthog. It can be removed

tombeckenham:
Closes #849

We don't use Langfuse anymore — PostHog is the only observability backend. This PR removes Langfuse entirely and, per discussion, replaces the bespoke LLM-analytics bridge with TanStack AI's official OpenTelemetry middleware exporting directly to PostHog. Rebased onto latest `main` (#906 frames→shots, styles work, per-call LLM cost tracking).

## Part 1 — Langfuse removal

- Deleted `src/lib/observability/langfuse.ts`, `tracer.ts`, and the custom OTel `gen_ai` span layer — its only exporter was the `LangfuseSpanProcessor`
- Deleted the Langfuse prompt scripts (`pull-prompts`, `upload-all-prompts`, `list-langfuse-prompts`, `print-schemas`); `workflow-prompts.ts` is now the single source of truth for prompts (local-only resolution, no two-tier fallback)
- Dropped all `promptReference` plumbing — this also removes a redundant second prompt fetch per durable LLM step
- Removed the `record-workflow-trace` step and the bespoke fal image/music/motion span observations that only ever exported to Langfuse (these are restored on the official path in Part 3)
- Removed `@langfuse/*` deps and `LANGFUSE_*` env wiring (`.env.example`, `worker-configuration.d.ts` via clean `cf:typegen`, `setup.ts`, deploy workflow secrets)
- Removed 4 response schemas + 2 schema exports that existed only for Langfuse prompt-config generation — knip output is now strictly better than main's baseline

## Part 2 — LLM analytics via TanStack AI OTel → PostHog

`chat()` calls now pass `middleware: aiObservabilityMiddleware(...)` (`src/lib/observability/ai-otel.ts`), which wraps TanStack AI's official `otelMiddleware` and exports `gen_ai.*` semconv spans straight to PostHog's OTLP AI endpoint (`/i/v0/ai/otel`). PostHog converts them into `$ai_generation` events server-side.

Two findings made this the right architecture:

- **PostHog now resolves `distinct_id` per-span** — the `posthog.distinct_id` span attribute beats Resource attributes (`posthog/rust/capture/src/otel/identity.rs`, explicitly for multi-tenant serverless). The old blocker that forced the custom `ai-event-bridge` is gone, so the bridge and the `@tanstack/ai-event-client` dep are deleted.
- **The PostHog span processor is vendored (~30 lines)** instead of adding `@posthog/ai`, which hard-depends on the OpenAI/Anthropic/Google/LangChain SDKs.

Net win: richer data than before — per-iteration generation spans with token usage, real `$ai_trace_id` hierarchy (`$ai_span` root + `$ai_generation` children), input/output capture, and `$ai_session_id` / `$ai_span_name` / `$ai_tags` passed through as span attributes.

The tracer provider is private (never registered globally), so the Cloudflare→PostHog logs connector / tail-worker forwarder can't double-export these spans.

## Part 3 — Upgrade @tanstack/ai 0.34 + media OTel spans

Bumped `@tanstack/ai` 0.32→0.34, `ai-fal` 0.9.0→0.9.2, `ai-openrouter` 0.14.0→0.14.2. 0.33's "Media Activity Observability" makes `otelMiddleware` emit `gen_ai.*` spans for `generateImage` / `generateAudio` / `generateVideo`, and 0.34's `otelMiddleware` returns `GenerationMiddleware & ChatMiddleware` so a single instance serves both chat and media.

- `aiObservabilityMiddleware` now returns the unified type and is threaded through the image/music/motion generators via an optional `observability` field on each `*Options`.
- This **restores the fal image/music/motion span observability** Part 1 had to drop — now on the official path, exporting to PostHog (`$ai_generation` with token/cost usage, per-user `distinct_id`, `$ai_session_id`, `$ai_span_name`, `$ai_tags`).
- No behavior change where PostHog is unconfigured (tests/e2e): the helper returns `[]`.
- Ride-along fixes: 0.33 corrects usage dropped on the structured-output fallback path (helps `onFinish` cost capture); 0.34 fixes null-for-optional-field rejection in structured output.

## Verification

- `bun typecheck` ✓ · `bun lint` ✓ (0 errors, 22 pre-existing warnings) · 1498 unit tests ✓ · `bun dead-code` ✓

## Follow-ups (manual, post-merge)

- Delete `LANGFUSE_*` from Doppler configs (dev/dev_test/prd/stg), GitHub repo secrets, and the deployed prod worker's secrets; revoke the Langfuse project keys
- `scripts/setup-prod.ts` still contains a Langfuse setup flow (a main-only file this branch never touched); strip it for coherence
- First live generation after deploy: confirm `$ai_generation` shows up in PostHog LLM Analytics with the correct user, including image/audio/video spans


tombeckenham:
Extra motivation from prod (PostHog logs, `openstory-prd`): this PR also fixes a **chronic global-scope error** that Langfuse init has been throwing on every isolate cold-start.

**Error (~300/day, all week, non-fatal):**
```
Error: Disallowed operation called within global scope. Asynchronous I/O
(ex: fetch() or connect()), setting a timeout, and generating random values
are not allowed within global scope. To fix this error, perform this operation
within a handler.
```

**Root cause:** `src/instrumentation.ts` (first import in `src/server.ts`, so it runs at Worker **global scope**) calls `initTracing()`, which constructs `new LangfuseSpanProcessor({ exportMode: batched })`. That extends OTel `BatchSpanProcessor`, whose batching pipeline schedules a flush `setTimeout` (`@opentelemetry/sdk-trace-base/.../BatchSpanProcessorBase.js:195`). Scheduling a timer in global scope is forbidden on Cloudflare Workers → the throw.

**Why the symptoms look odd:**
- `cf.outcome: ok` / non-fatal — thrown during isolate startup, logged, but the cold-starting request still completes; tracing just silently degrades.
- Attributed to `/sequences/new` (173), `/sequences` (5), `/gallery` (1) in a 12h sample — **not** route-specific; the error is stamped with whatever request triggered the isolate cold-start, so the highest-traffic pages dominate. Volume ≈ cold-isolate boots/day.
- Never caught locally — `initTracing()` early-returns unless both `LANGFUSE_PUBLIC_KEY` + `LANGFUSE_SECRET_KEY` are set (`isLangfuseEnabled`), which only holds in prod, so `bun dev`/e2e never build the processor.

**This PR resolves it** by deleting `langfuse.ts`/`initTracing()` entirely. Worth confirming the replacement path stays clean: the new `aiObservabilityMiddleware` / vendored PostHog span processor must not schedule any timer or I/O at module/global scope — the PR notes the tracer provider is private and never registered globally, and if the processor is constructed lazily inside the request-scoped middleware (not at import time) we stay clear of the same trap.

Post-merge: watch `openstory-prd` logs for `"Disallowed operation called within global scope"` — it should drop to zero.

Issue #885, opened 2026-06-11
Simplify local setup


tombeckenham:
I want to simplify local setup as much as possible. Cut down scripts in package.json the not necessary, and reduce the setup script substantially. The user should almost be able to run by bun install and bun dev. Have a good think how it could be simplfied further.

Include a [Deploy on Cloudflare](https://developers.cloudflare.com/workers/platform/deploy-buttons/) button

PR #918, opened 2026-06-12
feat(storage): store origin-relative media URLs so deployments need no VITE_APP_URL

### LINKED ISSUE #894: Store origin-relative media URLs so deployments need no VITE_APP_URL
## Problem

`getPublicUrl()` (src/lib/storage/buckets.ts) bakes an **absolute origin** into every media URL at write time: `${VITE_APP_URL}/r2/<key>` in local-serving mode, `https://${R2_PUBLIC_STORAGE_DOMAIN}/...` in CDN mode. Those absolute URLs are persisted into DB rows (frame imageUrl/videoUrl, sheets, library media, …), including from inside workflow isolates where no request context exists.

Consequences:

1. **`VITE_APP_URL` is a hard runtime requirement.** Without it (and no R2 domain), `getPublicStorageBase()` throws and all media reads/writes 500. This forces the Deploy-to-Cloudflare button flow (#885 / PR #893) to prompt for the deployment URL *before* the user has deployed — circular and confusing. It's the only hard dependency: auth derives its base URL from the request, and the deployment classifier/marketing fallbacks degrade gracefully.
2. **Domain migration silently breaks old media.** Moving a deployment (workers.dev → custom domain, or rotating the R2 public domain) leaves every previously generated row pointing at the old origin.

## Proposal

Store **origin-relative URLs** (`/r2/<key>`) — or better, just the bucket+key — and resolve to a displayable URL at read time:

- Browsers resolve relative URLs against the current origin automatically; the worker's own `/r2/$` route (src/routes/r2.$.ts) serves them. The page rendering a frame *is* the origin, so no config needed.
- Workflows stop needing to know any URL — they persist the path only.
- The few server-side consumers that genuinely need an absolute URL absolutize at the moment of use:
  - `external-url.ts` (making locally-served URLs fetchable by fal via fal.storage.upload / data-URIs) — detection actually gets simpler: `startsWith('/r2/')` instead of comparing against `VITE_APP_URL`.
  - `image-crop.ts` (`cdn-cgi/image/` wrapping) — derive from request origin or R2 public domain.
  - Sequence export / safe-fetch paths — audit `getPublicUrl` callers: storage-cloudflare.ts, copy-sequence-elements.ts, promote-temp-elements.ts, create-library-location.ts, create-library-talent.ts, safe-fetch.ts, sequence-exports.ts.
- CDN mode (`R2_PUBLIC_STORAGE_DOMAIN`) becomes a read-time concern: prepend the CDN origin when building display URLs instead of baking it into rows.

## Migration

Existing rows contain absolute URLs. Either:
- **Tolerant readers** (preferred): treat stored values as either absolute (legacy) or relative; new writes are relative-only. No data migration, old rows keep working on their original domain.
- Or a one-time normalization migration stripping known origins down to `/r2/<key>` (beware the D1 table-rebuild trap — this would be UPDATEs only, so safe).

## Done when

- `getPublicStorageBase()`'s `VITE_APP_URL` throw is gone; a worker deployed with **zero** URL config serves media correctly.
- `VITE_APP_URL` is removed from `.dev.vars.example` / the deploy-button prompts (`cloudflare.bindings` in package.json).
- Changing the serving domain does not break previously generated media.
- e2e fixtures/full pipeline pass (recorded fixtures reference locally-served URLs — re-record if needed).

Context: came out of #885 (PR #893) while wiring the Deploy to Cloudflare button — see the PR discussion for the rejected alternative (a self-configuring `deploy` script that `wrangler secret put`s the URL post-deploy).

tombeckenham:
## What

`getPublicUrl()` no longer bakes an absolute origin into persisted media rows.
It returns origin-relative `/r2/<bucket>/<path>`, and consumers that genuinely
need an absolute URL resolve it at the moment of use. Browsers resolve relative
URLs against whatever origin is serving the page, so:

- A worker deployed with **zero URL config** serves its own media — `VITE_APP_URL`
  is no longer a hard runtime requirement (the `getPublicStorageBase()` throw is
  gone). This unblocks the Deploy-to-Cloudflare button flow (#885 / PR #893),
  which previously had to prompt for the URL before the user had even deployed.
- **Changing the serving domain** (workers.dev → custom domain, or rotating the
  R2 public domain) no longer breaks previously generated media.

## How

- **`buckets.ts`**: drop `getPublicStorageBase()` and its `VITE_APP_URL` throw.
  Add `r2KeyFromUrl()` (extracts the object key; tolerant of legacy absolute
  `…/r2/<key>` rows) and `toCdnUrl()` (read-time absolutization against
  `R2_PUBLIC_STORAGE_DOMAIN`).
- **`/r2/$` route** now serves in every mode: streams from the R2 binding when
  no CDN domain is set, 302-redirects to the CDN domain when one is configured
  (so bytes are served/cached at the R2 edge, and rotating the domain only
  changes where the redirect points).
- **Server-side consumers** resolve at point of use: the fal/vision shim
  (`external-url.ts`) reads bytes straight from the R2 binding (new
  `readStorageObject` in `#storage`) instead of HTTP-fetching `VITE_APP_URL`;
  `image-crop` reads the PNG header from the binding; `image-compress`/`image-crop`
  absolutize transform URLs via the CDN domain.
- **`mediaUrlSchema`** (new) replaces `z.string().url()` on the ~12 validators
  that accept stored/relative URLs — `z.url()` rejects relative paths.
- **`AppImage`** absolutizes relative srcs via `VITE_APP_URL` for Cloudflare
  image transforms (the only origin SSR and client agree on, avoiding hydration
  mismatch); falls back to a plain `` without it.
- **aimock** fal normalization collapses bare `/r2/` and local `cdn-cgi`
  origins, so existing e2e fixtures match without re-recording.
- Remove `VITE_APP_URL` from `.dev.vars.example` and the deploy-button prompts
  (`cloudflare.bindings` in `package.json`).

## Migration: none needed

No schema change (`bun db:generate` → "No schema changes"; the `libraries.ts`
edit is a Zod validator on a JSONB-stored type, not a column) and **no data
migration**. Readers are tolerant: `r2KeyFromUrl()` accepts both the new
relative form (`/r2/<key>`) and legacy absolute rows that baked an origin in
front of the `/r2/` route. Legacy **CDN-domain** rows (e.g.
`https://storage.openstory.so/...`) aren't rewritten — they have no `/r2/`
prefix, so they pass through untouched and stay fetchable on their original
domain. New writes are relative-only; old rows keep working in place. (A
one-time normalization migration was considered and rejected as unnecessary —
it would be UPDATE-only and safe, but the tolerant-reader path makes it moot.)

## Testing

- typecheck, lint (0 errors), knip, format clean.
- 1366 unit tests pass, incl. new `buckets.test.ts` coverage for the URL helpers.
- Full-pipeline e2e (`test:e2e:full`, workflows + aimock): passes with existing
  fixtures (no re-record).
- `bun run build` passes; built `getPublicStorageBase` throw is gone.

Closes #894


tombeckenham:
Pushed a follow-up fixing the share-by-URL regression flagged in review: stored URLs are origin-relative, but two egress points handed a media URL to something off-origin (the public API v1 `buildSequenceState` response, and the "Copy latest export URL" action) and were emitting the bare `/r2/...` path.

New `toShareableUrl(url, origin)` helper absolutizes at egress — CDN domain when configured, else the request/window origin; already-absolute external/legacy URLs pass through. Applied in the v1 state builder (origin threaded through both routes) and the theatre copy handler. In-app download/playback were already fine (browser resolves relative URLs). Added unit coverage for the helper and the API mapping (relative→absolute and passthrough cases).

PR #923, opened 2026-06-15
Add drag-and-drop scene reordering

tombeckenham:
## Summary

Adds drag-and-drop reordering of scenes (frames) in the desktop scenes sidebar.

- **UI**: `SceneList` now wraps frames in dnd-kit's `DndContext` + `SortableContext` (vertical strategy). New `SortableSceneItem` gives each card a left-gutter grip handle that is the sole drag activator — clicking the card body still selects the scene, and the handle is keyboard-operable (focus + arrow keys) per the WAI-ARIA reorder pattern. Reordering auto-disables when there's ≤1 frame.
- **Data**: new `useReorderFrames()` mutation with optimistic cache reordering (instant UI), rollback on error, and invalidate-on-settle. Reuses the existing `reorderFramesFn` server endpoint.
- **Bug fix**: `scopedDb.frames.reorder` was dormant (no UI caller, no tests) and would have crashed on any real reorder — SQLite checks the `(sequence_id, order_index)` unique index immediately, so the single-pass batch collided. Now two-phase: park every frame in a distinct negative slot, then assign final indices.

## Dependencies

Adds `@dnd-kit/core`, `@dnd-kit/sortable`, `@dnd-kit/utilities` (React 19 compatible, accessible, tree-shakeable).

## Testing

- New `src/lib/db/scoped/frames-reorder.test.ts` — real in-memory libsql + actual migrations (so the real unique index is present). Covers adjacent swap, full reverse, rotation, and empty no-op. These specifically exercise the cases the old single-pass batch would have failed.
- `bun typecheck` and `bun lint` clean.

## Notes

- Covers the **desktop** scene sidebar. The mobile drawer renders `SceneListItem` directly and is not yet draggable.


tombeckenham:
this was just a test

PR #1008, opened 2026-06-30
Make public talent read-only

### LINKED ISSUE #666: Deleting public/system talent fails silently with "Talent not found or failed to delete"
## Summary

On `/talent/$id`, clicking delete on a public/system talent (one seeded via `DEFAULT_SYSTEM_TALENT` and owned by the system team) returns `Talent not found or failed to delete` and leaves the UI stuck — no toast, no navigation, no visible feedback.

## Reproduction

1. Sign in to a normal team.
2. Open the talent library — public/system talents appear in the list (because list queries OR on `isPublic: true`).
3. Open one and click the trash icon → confirm.
4. Server returns 500 with `Talent not found or failed to delete`. UI does nothing.

## Root cause

- List/getById include public talents:
  - `src/lib/db/scoped/talent.ts:24-27` — `or(eq(talent.teamId, teamId), eq(talent.isPublic, true))`
  - `src/lib/db/scoped/talent.ts:167-173` — `getById` uses `OR: [{ teamId }, { isPublic: true }]`
- But delete is strictly team-scoped:
  - `src/lib/db/scoped/talent.ts:263-269` — `where(and(eq(talent.id, talentId), eq(talent.teamId, teamId)))`
  - For a system-team talent, `rowsAffected === 0` → `deleteTalentFn` throws `Talent not found or failed to delete` (`src/functions/talent.ts:182-194`).
- UI hides nothing for non-owned talents and has no error surface:
  - `src/routes/_protected/talent/$id.tsx:152-159` — delete button rendered unconditionally
  - `src/routes/_protected/talent/$id.tsx:69-75` — `handleDelete` awaits `mutateAsync` then navigates; on error the navigate never fires
  - `src/hooks/use-talent.ts:89-98` — `useDeleteTalent` has no `onError`

## Production log

```
{
  "name": "deleteTalentFn",
  "method": "POST",
  "error": { "code": "UNKNOWN", "message": "Talent not found or failed to delete" },
  "durationMs": 127
}
```

## Proposed fix

1. Hide edit/delete (and probably "Generate Sheet" too) on `/talent/$id` when `talent.teamId !== currentTeamId` or `talent.isPublic === true` — system templates should be read-only for non-owners.
2. Add `onError` toast in `useDeleteTalent` (and the other talent mutations) so future server-side failures are visible instead of silently leaving the page in a stuck state.
3. Optional: tighten the server error to distinguish "not yours" vs "not found" — return 403 for the public-talent case.

## Out of scope

If we want users to hide public talents they don't care about, that's a separate "hidden talents" feature (likely a per-team join table) and should be a different issue.

AdemBenAbdallah:
## Summary

Closes #666

- Hide owner-only actions on public/system talent detail pages for non-owning teams
- Gate edit/delete/favorite/generate-sheet/set-default controls behind team ownership
- Surface delete failures with a toast instead of silently leaving the page stuck
- Return clearer server-side delete errors for not-found, forbidden, and failed-delete cases

## Testing

- bun typecheck
- bun format:check
- bun lint: 0 errors; existing unrelated warnings remain

Pre-commit also passed dead-code, format, lint, and typecheck for the staged files.

### DIFF
--- a/src/functions/talent.ts
+++ b/src/functions/talent.ts
@@ -6,7 +6,7 @@ import {
   getPublicTalentWithRelations,
   listPublicTalent,
 } from '@/lib/db/scoped';
-import type { Talent, TalentWithSheets } from '@/lib/db/schema';
+import type { TalentWithSheets } from '@/lib/db/schema';
 import { ulidSchema } from '@/lib/schemas/id.schemas';
 import {
   createTalentSchema,
@@ -34,23 +34,6 @@ const talentIdSchema = z.object({ talentId: ulidSchema });
 const sheetIdSchema = z.object({ sheetId: ulidSchema });
 const mediaIdSchema = z.object({ mediaId: ulidSchema });
 const characterIdSchema = z.object({ characterId: ulidSchema });
-
-/**
- * Verify a talent record belongs to the given team, throwing if not found.
- * Uses scopedDb which is already team-scoped.
- */
-async function requireTalentOwnership(
-  scopedDb: {
-    talent: { getById: (id: string) => Promise<Talent | undefined> };
-  },
-  talentId: string
-): Promise<Talent> {
-  const record = await scopedDb.talent.getById(talentId);
-  if (!record) {
-    throw new Error('Talent not found');
-  }
-  return record;
-}
 
 // List Talent
 
@@ -175,8 +158,6 @@ export const createTalentSheetFn = createServerFn({ method: 'POST' })
   .middleware([authWithTeamMiddleware])
   .inputValidator(zodValidator(createTalentSheetSchema))
   .handler(async ({ context, data }) => {
-    await requireTalentOwnership(context.scopedDb, data.talentId);
-
     return context.scopedDb.talent.sheets.create({
       talentId: data.talentId,
       name: data.name,
@@ -204,8 +185,6 @@ export const deleteTalentSheetFn = createServerFn({ method: 'POST' })
       throw new Error('Sheet not found');
     }
 
-    await requireTalentOwnership(context.scopedDb, sheet.talentId);
-
     const deleted = await context.scopedDb.talent.sheets.delete(data.sheetId);
     if (!deleted) {
       throw new Error('Failed to delete sheet');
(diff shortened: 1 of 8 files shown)


src/functions/talent.ts
```diff
@@ -141,11 +141,19 @@ export const deleteTalentFn = createServerFn({ method: 'POST' })
   .middleware([authWithTeamMiddleware])
   .inputValidator(zodValidator(talentIdSchema))
   .handler(async ({ context, data }) => {
+    const talentRecord = await context.scopedDb.talent.getById(data.talentId);
+    if (!talentRecord) {
+      throw new Error('Talent not found');
+    }
+    if (talentRecord.isPublic || talentRecord.teamId !== context.teamId) {
```
tombeckenham: **[critical]** Good guard, but it is the *only* server function in this PR that rejects `isPublic`. The same writable check must be applied to `updateTalentFn`, every handler calling `requireTalentOwnership`, `generateTalentSheetFn`, and `talent-sheet-variants` promote/discard/undiscard. Extract `assertTalentWritable` and use it for all mutations.

src/routes/_app/talent/$id.tsx
```diff
@@ -69,12 +77,21 @@ function TalentDetailPage() {
     generateSheet.mutate({ talentId: talent.id });
   };
 
-  const handleDelete = async () => {
+  const canManageTalent = Boolean(
```
tombeckenham: **[important]** `canManageTalent` correctly gates the UI, but UI-only enforcement is bypassable via direct server-fn calls. Do not merge until the server-side audit in the review body is addressed.

AdemBenAbdallah:
Thanks for rebasing this one too. I noticed the latest `Test` and `Deploy Cloudflare` workflow runs failed instantly with no jobs/logs attached, and GitHub currently reports no checks on the PR branch.

This does not look like a code/test failure because no jobs were created. Could you rerun the checks when you get a chance?

tombeckenham:
The ci tests were failing as they (wrongly) were set up to need secrets.

AdemBenAbdallah:
that makes sense.

tombeckenham (approved):
Good one to find. Thanks for helping lock that down

PR #1014, opened 2026-07-01
test(e2e): select style card by aria-label in full-sequence pipeline (#821)

### LINKED ISSUE #821: Flaky full-sequence e2e: first style-grid click can land on the 'More Options' tile before style tiles render
Found while verifying #815 locally (the failure is pre-existing and unrelated to that change — CI passes this step on the same code).

## Symptom

`e2e/tests/full-sequence.spec.ts` step 2 selects a style with:

```ts
await page
  .getByRole('grid', { name: 'Style selection' })
  .getByRole('button')
  .first()
  .click();
```

On a slow start (cold vite dev in a fresh worktree, no `.vite` cache), this click lands on the **"More Options" tile** instead of a style card. That opens the Visual Style dialog; the subsequent `scriptTextarea.fill(script)` then gets focus-trapped into the dialog's search input (the page snapshot at failure shows the whole script inside the dialog's `Search` textbox and "No styles found"), and the test dies at:

```
Error: expect(locator).toBeEnabled() failed
Locator: getByRole('button', { name: /Enhance Script/i })
Error: element(s) not found   ← page is aria-hidden behind the modal
```

Reproduced deterministically twice on a cold local worktree; CI's built-server full-pipeline job is fast enough to never hit it.

## Mechanism

In `src/components/style/style-selector.tsx`, the grid renders `visibleStyles` (sliced by a `visibleCount` derived from the measured container width) plus a trailing "More Options" button gated only on `!loading`. In the window where `loading` is false but `visibleCount` is still 0 (container width not yet measured / layout not settled), the **only** `role=button` in the grid is "More Options".

## Suggested fix

Make the test target a style card explicitly and wait for one to exist, e.g.:

```ts
const firstCard = page
  .getByRole('grid', { name: 'Style selection' })
  .getByRole('button', { name: /^Select .* style$/ })
  .first();
await expect(firstCard).toBeVisible();
await firstCard.click();
```

(Style cards have `aria-label="Select <name> style"`; the More Options tile is labelled `View all N styles`.)

Optionally also harden the component: don't render the More Options tile until `visibleCount > 0`.

agualis:
## Related Issue

Closes #821

## Summary of Changes

The full-sequence e2e selected a style by clicking the grid's first button. On a cold start that button can briefly be the trailing "More Options" tile (rendered before the style cards settle), which opens the style dialog and traps the later script `.fill()` inside the modal — the run then fails at "Enhance Script".

This targets a style card explicitly by its `Select <name> style` aria-label and waits for it before clicking, so the click can never land on the "More Options" tile. Robust regardless of render timing. Test-only change; no product code touched.

Verified locally: `bun test:e2e:full` passes end-to-end, and `bun lint` / `bun format:check` / `bun typecheck` / `bun run test` are all green.

## Risk Assessment

- [x] Low

## Additional Notes

👋 First-time contributor here. I really like the project and the stack, and I'll have some free time coming up that I'd love to put toward contributing more.

I also noticed @AdemBenAbdallah recently started jumping in around the same time, great to see! Just flagging that I'm here in a team spirit, not to compete: happy to coordinate so we don't step on each other's toes or duplicate work. Feel free to tag some "good first issues", or point us to where you most need help.


tombeckenham:
Thank you. I've just been looking at these tests. Much appreciated!

tombeckenham:
I ended up having to close this I had to apply to fix on another PR. Please feel free to raise anything else to get contributions into the repo. It's great to have help

PR #1015, opened 2026-07-01
Respect Fal proxy for per-request script clients

### LINKED ISSUE #890: Per-request fal clients bypass FAL_PROXY_URL (external-url.ts, verify-fal-costs.ts)
## Context

Found while triaging the knip dead-export backlog (#786).

`configureFalProxyFromEnv()` in `src/lib/ai/fal-config.ts` routes server-side fal.ai traffic through a proxy when `FAL_PROXY_URL` is set, by monkey-patching `fal.config` on the singleton client. The project used to also export a local `createFalClient` wrapper so per-request clients picked up the same proxy middleware — but nothing imported that wrapper (it was dead code and has been removed in #786).

## Problem

Two call sites build per-request fal clients directly via `@fal-ai/client`'s `createFalClient(...)`, which returns an independent client whose config closure the monkey-patch can't touch — so these bypass `FAL_PROXY_URL` entirely:

- `src/lib/storage/external-url.ts` (`createFalClient({ credentials: getEnv().FAL_KEY })` for `fal.storage.upload`)
- `scripts/verify-fal-costs.ts`

If the proxy path matters for those flows (cost attribution, egress control, observability), they should compose the same request middleware (or reintroduce a shared wrapper that does). If the proxy is only needed for the adapter/singleton path, document that the per-request clients intentionally go direct.

## Acceptance

- Decide whether `fal.storage.upload` (and the verify-costs script) must respect `FAL_PROXY_URL`.
- Either wire the proxy middleware into those per-request clients, or document the intentional bypass next to `configureFalProxyFromEnv`.

AdemBenAbdallah:
## Summary
- add a proxy-aware per-request Fal client factory that composes the existing FAL_PROXY_URL middleware
- use it in the fal cost verification script so script clients do not bypass the proxy
- keep fal storage uploads intentionally direct and document that exception from the shared helper
- add unit coverage for proxy rewriting, middleware composition, and non-Fal passthrough

Closes #890

## Validation
- bun run test src/lib/ai/fal-config.test.ts
- bun typecheck
- bun format:check
- bun lint
- pre-commit: dead-code, format, lint, typecheck

### DIFF
--- a/scripts/verify-fal-costs.ts
+++ b/scripts/verify-fal-costs.ts
@@ -18,6 +18,7 @@
  *   bun scripts/verify-fal-costs.ts --compare            # fetch usage for existing results
  */
 
+import { createProxiedFalClient } from '@/lib/ai/fal-config';
 import { estimateFalCost } from '@/lib/ai/fal-cost';
 import {
   AUDIO_MODELS,
@@ -29,7 +30,6 @@ import {
 import { buildModelInput } from '@/lib/motion/build-model-input';
 import { snapDuration } from '@/lib/motion/motion-generation';
 import { typedEntries } from '@/lib/utils/typed-object';
-import { createFalClient } from '@fal-ai/client';
 import { mkdir, readFile, stat, writeFile } from 'node:fs/promises';
 import path from 'node:path';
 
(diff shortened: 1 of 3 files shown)


tombeckenham:
I think that issue was created by an agent during an automated review. I think the "alternative" suggestion was the right approach - comment why we're not proxying. The fal costs script exists to get real fal costs. I'm going to close this pr, and open a new one to change that comment and close this issue

PR #1027, opened 2026-07-02
Keep style public flag server-managed

### LINKED ISSUE #869: Style copies in user teams can show as duplicate public templates (isPublic client-settable)
## Symptom

A style template can appear **twice** in the gallery — once as the legit System Templates row, and once as a copy living in a *user's own team* that is flagged `is_public=1` (and `is_template=1`).

Concrete case found in local dev (now manually cleaned up): a byte-identical copy of the system "Product Ad" style existed in a user team (`is_template=1, is_public=1`) with 4 real sequences referencing it. The style list query surfaces both because it is:

```ts
// src/lib/db/scoped/styles.ts:30
.where(or(eq(styles.teamId, teamId), eq(styles.isPublic, true)))
```

so any `isPublic=true` style — including one in your own team — renders next to the system templates.

## Confirmed finding

`isPublic` is **not** in `SERVER_MANAGED_COLUMNS`, unlike `isTemplate`:

```ts
// src/lib/schemas/style.schemas.ts:20
const SERVER_MANAGED_COLUMNS = {
  id, teamId, createdBy, createdAt, updatedAt,
  usageCount, version, isTemplate, sortOrder,
} as const;
// createStyleSchema = createInsertSchema(styles, …).omit(SERVER_MANAGED_COLUMNS)
```

So a client can set `isPublic: true` on `createStyleFn`, publishing a team-local style into everyone's public gallery. This is both a duplicate-UX bug and a data-integrity / leakage concern (one team's style becomes globally visible).

## Open question (needs root-cause)

The found copy also had `is_template=1`, which *is* protected by the create schema — so it was not created via `createStyleFn`. Likely an older/other "copy template into team" path or a raw `db.insert(styles)` that duplicates a template row wholesale and retains its flags. Needs tracing: how does a team end up with a local copy of a system template at all (sequence-creation "use this style" flow?), and why does the copy keep `is_public`/`is_template`?

## Suggested fix

1. Add `isPublic: true` to `SERVER_MANAGED_COLUMNS` so clients cannot self-publish styles (publishing should be a deliberate server-gated action).
2. Find the copy-template-into-team path and ensure copies are created with `isPublic=false, isTemplate=false`.
3. Consider a one-off prod audit: `SELECT … FROM styles WHERE is_public=1 AND team_id != <system team>` to find any leaked public styles.

Found while implementing #866 (Grok Imagine Video 1.5).

AdemBenAbdallah:
## Summary
- make `isPublic` server-managed in style create/update schemas, matching `isTemplate`
- prevent client-submitted style inputs from publishing team-owned styles into the public catalogue
- add schema regression tests covering create/update stripping of public/template and other server-managed fields

## Follow-up hardening (maintainer commit)
Review of the original fix found the same issue class elsewhere plus two enforcement gaps, addressed in `fa0fa9e0`:

- **Talent had the identical hole, and it was live**: `updateTalentSchema` didn't omit `isPublic`, and `updateTalentFn` → `scopedDb.talent.update` spreads client data into `.set()`, so any team member could publish their own talent into the anonymous public catalogue (`getPublicTalentFn`). `isPublic`/`isTemplate` are now stripped from both talent schemas.
- **Schema omits alone weren't enough.** The `Omit<>` parameter types on the scoped write methods are compile-time only — TypeScript's excess-property check applies just to fresh object literals, and drizzle's `.values()`/`.set()` writes any key that matches a real table column. The scoped styles/talent/location write methods now scrub server-managed columns at runtime (`stripServerManagedColumns`) before spreading.
- **One definition drives all layers**: the `SERVER_MANAGED_*` column lists are exported consts, shared by the Zod `.omit()`, the scoped-method parameter types, and the runtime scrub.
- New `typedFromEntries` util (inverse of `typedEntries`); regression tests pin the full omit lists and prove the runtime scrub drops a smuggled `isPublic` from a non-literal object on both create and update paths.

Locations were already safe at the schema level (inline allow-list validators) — they get the type + runtime defense for consistency.

Closes #869

## Validation
- `bun run test` (full suite, 1636 tests)
- `bun typecheck`
- `bun lint` / `bun format:check` / `bun dead-code`

### DIFF
--- a/src/lib/db/scoped/location-library.ts
+++ b/src/lib/db/scoped/location-library.ts
@@ -6,11 +6,32 @@
 import { eq, ilike, and, inArray, or } from 'drizzle-orm';
 import type { Database } from '@/lib/db/client';
 import { locationLibrary, locationSheets } from '@/lib/db/schema';
+import { stripServerManagedColumns } from './server-managed';
 import type {
   LibraryLocation,
   NewLibraryLocation,
   NewLocationSheet,
 } from '@/lib/db/schema';
+
+// Columns callers must never set through the scoped write methods.
+// id/teamId/createdBy/createdAt/updatedAt are injected here or by the
+// database, and the public/template flags are admin/seeder-only — the seeder
+// inserts via raw drizzle. (Locations have no shared Zod schema module; the
+// server-fn validators are inline allow-lists, so this is the second layer
+// of the same defense — excluded from the parameter types AND scrubbed at
+// runtime, since a non-literal object can carry extra keys past an Omit<>
+// parameter and drizzle writes any key that matches a table column.)
+const SERVER_MANAGED_LOCATION_COLUMNS = {
+  id: true,
+  teamId: true,
+  createdBy: true,
+  createdAt: true,
+  updatedAt: true,
+  isPublic: true,
+  isTemplate: true,
+} as const;
+
+type ServerManagedLocationColumn = keyof typeof SERVER_MANAGED_LOCATION_COLUMNS;
 
 /**
  * Public (anonymous) location-library reads. Takes no team scope at all, so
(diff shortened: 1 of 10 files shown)


tombeckenham (commented):
Ok, did we miss something when we look at this previously?

AdemBenAbdallah:
I don’t think this was fixed in the earlier PRs -  #1008 was talent-specific. For styles, the leak path is still open because `createStyleSchema` lets client input set `isPublic`, and `styles.create()` spreads that into the insert.

I also checked for a style copy/clone path and only found the intentional system-template seeder, so I think the schema fix is still the right one.

tombeckenham (approved):
Thanks for finding this. Totally missed this issue. I also strengthed runtime checks to make sure public couldn't slip through. But... I am at some point considering public styles

PR #1041, opened 2026-07-08
Reject music-only sequence creation

### LINKED ISSUE #823: Music never generates when music:true but motion:false (silently skipped)
## Summary

When a sequence requests **music but not motion** (`autoGenerateMusic: true`, `autoGenerateMotion: false`), music is **never generated** and no error is surfaced. The sequence reaches a terminal `completed` state with `musicStatus` stuck at `pending`.

Found during review of the public `/api/v1` one-shot API (PR #804), where `motion` and `music` are independent caller-supplied flags — so `{ "motion": false, "music": true }` is a perfectly valid request that silently does nothing for music.

## Root cause

In `src/lib/workflows/analyze-script-workflow.ts`, `shouldGenerateMusic` is computed (line ~592) but is **only ever consumed inside the `if (shouldGenerateMotion)` block** (line ~600). Music is carried as part of the `motion-batch` child workflow payload (`includeMusic`, `audioModels`, `music`). There is no `else` / music-only path after the block closes (~line 686).

So when `shouldGenerateMotion` is false — because `autoGenerateMotion` is false, there's no primary video model, or there are no image URLs — the entire block is skipped and music generation never runs.

```ts
const shouldGenerateMotion =
  autoGenerateMotion && primaryVideoModel && imageUrls.length > 0;
const shouldGenerateMusic = Boolean(
  autoGenerateMusic && sequenceId &&
  completeScenes.some((s) => s.musicDesign?.presence && s.musicDesign.presence !== 'none')
);

if (shouldGenerateMotion) {
  // ... spawns motion-batch child with includeMusic: shouldGenerateMusic
}
// <-- no music-only branch here
```

## Impact

- **API callers:** `POST /api/v1/sequences` with `{ "motion": false, "music": true }` returns `202` and completes, but `music.status` in the status document never leaves `pending`. An agent that requested music and polls (incl. `?wait`) sees no music and no error.
- **UI:** likely reachable too if a sequence can be configured for music without motion.

## Suggested fixes (pick one)

1. **Add a music-only path:** when `shouldGenerateMusic && !shouldGenerateMotion`, trigger the music generation independently (e.g. a standalone music workflow / the existing music step) so the two flags are truly independent.
2. **Validate the combination:** reject `music && !motion` at the input layer with a clear `400` (e.g. in `apiCreateSequenceSchema` for the API, and the equivalent UI guard), so callers learn it's unsupported instead of getting a silent no-op.

Option 1 is the better long-term behaviour (the flags read as independent); option 2 is the cheap stopgap that at least makes it legible.

## Acceptance

- `{ motion: false, music: true }` either generates music (status reaches `completed`/`failed`) **or** is rejected with an explanatory error — never silently leaves `musicStatus: pending` on a `completed` sequence.

_Refs: `src/lib/workflows/analyze-script-workflow.ts` (~L590–686); surfaced via PR #804 review._

AdemBenAbdallah:
## Summary

This makes the create path reject `music: true` when `motion: false` instead of silently skipping music generation.

### Why

The current workflow only starts music generation inside the motion branch, so a music-only request becomes a silent no-op. This change makes that limitation explicit at validation time so callers get a clear error instead of a completed sequence with `musicStatus: pending`.

### What changed

- Added a `superRefine` guard to `apiCreateSequenceSchema`
- Added the same guard to `createSequenceSchema`
- Added regression tests for both schemas

### Notes

I kept this scoped to validation so we can make the current behavior legible first. If the preferred product direction is to support music-only generation, I can follow up with a standalone music path.

Closes #823

### DIFF
--- a/src/functions/__tests__/sequences.test.ts
+++ b/src/functions/__tests__/sequences.test.ts
@@ -17,6 +17,7 @@ import {
 import {
   assertModelNotAlreadyAdded,
   buildAddAudioMusicInput,
+  musicWithoutMotion,
   selectEligibleVideoShots,
   sumShotDurationsSeconds,
 } from '@/functions/sequences';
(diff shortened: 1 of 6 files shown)


tombeckenham (dismissed):

tombeckenham (approved):

tombeckenham (commented):
Thank you for this!

tombeckenham (approved):

PR #1058, opened 2026-07-17
fix: time out stalled Fal image requests

### LINKED ISSUE #826: Workflow image-generation steps have no timeout — a hung fal call stalls the pipeline silently
Found while test-rendering style sample videos through the real pipeline (#801).

## What happened
A `POST /api/v1/sequences` run (local dev, Workerd) stalled indefinitely: `CharacterSheetWorkflow`'s `generate-sheet-image` step and `LocationSheetWorkflow`'s reference generation both started their `nano_banana_2` fal calls and never returned — no response, no error, no retry — for 12+ minutes, while a parallel `flux_2_turbo` poster generation on the same run completed in ~4s.

Local explorer API (`/cdn-cgi/explorer/api/workflows/character-sheet-workflow/instances/...`) showed the step attempt with `start` set and `end: null`, status `running`, error `null`. The parent `analyze-script` workflow (and therefore the whole sequence) sat in `processing` with no progress and no `statusError`. A second identical run immediately after went straight through, so the hang looks like a transient fal/network condition — but nothing in our stack bounded it.

## Why it's a problem
- `generateImageWithProvider` → `@tanstack/ai-fal` adapter has **no timeout/abort**; a dead connection or stuck queue hangs forever.
- The workflow step never fails, so Cloudflare Workflows' step retry (which would likely have succeeded, as the retry run proved) never kicks in.
- Externally the sequence is indistinguishable from "slow generation" — API consumers long-poll `?wait=` forever; `statusError` stays null.

## Suggested fix
Wrap provider calls inside workflow steps (image gen at minimum; probably motion/music submit+poll too) with a generous `AbortSignal.timeout(...)` (e.g. 5–10 min for image generation) so a hung call becomes a step failure and the step's built-in retry can do its job.

## Repro pointers
- Step instance inspection: `GET /cdn-cgi/explorer/api/workflows/{name}/instances/{id}` in local dev.
- Manual unstick: `PATCH .../instances/{id}/status` with `{"action":"terminate"}`.

Refs: #801 (where this was hit), `src/lib/image/image-generation.ts`, `src/lib/workflows/character-sheet-workflow.ts`.

AdemBenAbdallah:
## Related Issue

Closes #826

## Summary of Changes

- Configure Fal image adapters with a timeout-aware fetch implementation.
- Abort requests that remain unresolved for more than ten minutes while preserving upstream cancellation signals.
- Let the existing Cloudflare Workflow step retry path handle the resulting request failure instead of leaving image generation running indefinitely.

## Risk Assessment

- [x] Low
- [ ] Medium
- [ ] High

## Additional Notes

The timeout is local to the existing image-generation module because it is currently the only consumer. The custom fetch is the documented configuration boundary exposed by the TanStack AI Fal adapter.

## Screenshots (if applicable)

Not applicable — reliability-only change.

### DIFF
--- a/src/lib/image/image-generation.ts
+++ b/src/lib/image/image-generation.ts
@@ -26,6 +26,7 @@ import { falImage } from '@tanstack/ai-fal';
 import { getLogger } from '@/lib/observability/logger';
 
 const logger = getLogger(['openstory', 'image', 'image-generation']);
+const IMAGE_GENERATION_TIMEOUT_MS = 10 * 60 * 1000;
 
 export type ImageGenerationParams = {
   model: TextToImageModel;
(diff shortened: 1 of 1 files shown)


tombeckenham:
Thanks for this — it correctly targets the hang class via the `falImage` custom-`fetch` option and matches the 10-minute budget we wanted for image generation.

We've since landed the same fix (and then a stronger follow-up) on main:

- **#1128** (`79680606`) — shared wall-clock `createDeadlineFetch` for image/music/motion so a hung connection *or* a never-finishing poll loop fails the workflow step, with unit tests and clear timeout messages
- **`da9cb2c6`** — bump `@tanstack/ai` to 0.44 and use native activity-level `timeout` on `generateImage` / media paths (TanStack/ai#981); deadline-fetch remains only for raw queue/status calls

This branch is ~200 commits behind, conflicts with the heavily refactored `image-generation.ts`, and its per-request timer is strictly weaker than the wall-clock / native timeout approach already on main.

Closing as **superseded**. #826 is resolved on main.

Issue #1062, opened 2026-07-22
10 bucks may not be enough


tombeckenham:
## Problem

Users report insufficient credits to generate a storyboard on their first attempt after signup.

## Root cause

**Not** an underfunded $10 grant. Preflight estimation treated `openai/gpt-image-2` as **$1/image**.

Fal's pricing API returns:
- `unit_price: 1`, `unit: "units"` (i.e. $1 **per billable unit**)
- Historical estimate: **~$0.22 per call** → ~0.22 units at high quality

Actual billing already uses `unitsBilled × unitPrice` correctly. Preflight assumed 1 unit per image → ~$14 storyboard estimate vs ~$3.14 real, blocking new users under a $10 grant.

## Fix

1. `bun scripts/update-fal-pricing.ts` now also pulls fal's `historical_api_price` estimate and stores `typicalUnitsPerCall = historical_cost / unit_price` in `fal-pricing-data.ts`.
2. `estimateFalCost` for image/flat models uses `typicalUnitsPerCall` (gpt-image-2 ≈ 0.22, not 1). Duration models stay parametric.
3. Signup grant stays **$10** (covers default storyboard ~$3.14 with headroom).
4. Regression tests: gpt-image-2 estimate < $1/image; grant ≥ default storyboard preflight.

## Out of scope

- No backfill for existing teams
- Motion-on first runs still cost more (video); users enabling motion may still need a top-up

Issue #1019, opened 2026-07-02
Scenes not listing


tombeckenham:
I'm getting an issue where on some sequences, the scenes aren't listing out. e.g. https://openstory.so/sequences/01KWGDCGQ3GTFF202MSBZ2QMNK/scenes

tombeckenham:
Root cause (confirmed via PostHog logs, service `openstory-prd`):

`getShotsFn` calls `scopedDb.shots.ensureAnchorFrames(shotRows)` with **every** shot of the sequence at once, and `ensureAnchorFrames` did a single unchunked `INSERT INTO frames …`. Each anchor row binds ~10 params, so a sequence with more than ~10 shots overflowed **D1's 100-bound-parameter ceiling** and the whole query threw:

```
serverFn "getShotsFn" failed: "UNKNOWN" "Failed query: insert into \"frames\" … (22+ value tuples) …"
```

When `getShotsFn` rejects, `useShotsBySequence` leaves `shots` `undefined`, and `SceneList` renders skeletons forever (it treats `undefined`/empty the same as loading) — so scenes never list. This only bit **long** sequences, which matches "on some sequences" (e.g. `01KWGDCGQ3GTFF202MSBZ2QMNK`). The React #419 SSR errors on the `/scenes` URL are the downstream symptom.

Fix:
- Chunk `ensureAnchorFrames` inserts (9 rows/chunk, ~90 params) and merge the RETURNING maps.
- Also chunked `getSelectedMotionByShots` (unchunked `inArray` on the same read path — would overflow at ~100 shots) at 90 ids/chunk.
- Regression test covering a shot count that spans multiple chunks.

Still open (separate, likely related to this same pathological mega-sequence, tracked here for context): "Worker exceeded memory limit" and "Disallowed operation called within global scope" errors in the same window — the global-scope one has no stack and appears fresh-isolate-wide, so it looks like a latent module-init issue rather than the `getShotsFn` path.

tombeckenham:
[...]

PR #1084, opened 2026-07-29
feat(#1069): DB-backed model pricing with observed unitsBilled medians

### LINKED ISSUE #1069: fal cost estimates for compute-seconds models are up to ~100x low (Grok Imagine, FLUX.2)
## Summary

Pre-flight cost estimates for fal models billed in **compute seconds** are wrong by up to ~100×, because the estimator falls back to a fixed `DEFAULT_COMPUTE_SECONDS = 3` when it has no real unit count.

Pricing has been hidden from the "Add a model" UI as an interim measure. It should come back as an opt-in display once estimates are trustworthy.

**Billing is not affected.** Actual charges use fal's reported `unitsBilled` (`falCostFromUnits`), which is exact. This is estimation and credit-gating only.

## The numbers

Observed in the sequence Image model dropdown for a 5-scene, 16:9 sequence:

| Model | Shown | Actual | Error |
| --- | --- | --- | --- |
| Grok Imagine Image Quality | `~$0.0026` | `~$0.25` | **~98× low** |
| FLUX.2 Dev | `~$0.03` | `~$0.086` | **~3.4× low** |

Grok's real price per fal's docs is $0.05/image at 1K ($0.07 at 2K). We estimated $0.00051.

The other 8 models in that dropdown are correct — they are billed in `images` or `megapixels`, where the unit count is derivable from the request.

## Root cause

`src/lib/ai/fal-cost.ts:28` — `DEFAULT_COMPUTE_SECONDS = 3`.

For `compute_seconds` models the estimator needs a seconds-per-image count. It reads `typicalUnitsPerCall` from `fal-pricing-data.ts`; when absent it assumes 3 compute-seconds. Grok needs ~294 to reach $0.05. FLUX.2 needs ~10. No single default can serve both — the spread is ~30×.

`typicalUnitsPerCall` is absent for a legitimate reason. `scripts/update-fal-pricing.ts` derives it from fal's `historical_api_price` endpoint, and that endpoint returns `total_cost: 0` for **all six** of our compute-seconds endpoints — fal has no usage history for them:

- `xai/grok-imagine-image/quality/text-to-image`
- `xai/grok-imagine-image/quality/edit`
- `fal-ai/flux-2`
- `fal-ai/flux-2/edit`
- `fal-ai/flux-2/turbo/edit`
- `fal-ai/hunyuan-image/v3/instruct/edit`

The stored `unitPrice` values are correct and verbatim from fal's pricing API. Only the unit **count** is missing.

## Blast radius

1. The "Add a model" estimate (now hidden).
2. **The pre-flight credit gate**, via the same `estimateImageCost`. These models under-gate, so a user near zero balance can start work they cannot pay for.
3. The four `/edit` endpoints above, wherever edit-flow costs are estimated.

## Proposed fix: derive unit counts from our own observed `unitsBilled`

No hand-curated constants. We already measure the exact quantity `typicalUnitsPerCall` is trying to approximate — fal returns it on every generation via the `x-fal-billable-units` header, surfaced by `@tanstack/ai-fal` as `usage.unitsBilled`, and we consume it at three sites (`image-generation.ts:182`, `music-generation.ts:206`, `motion-workflow.ts:642`).

Fal's `historical_api_price` is merely *their* aggregate of that same signal. Ours will be better: it reflects our resolutions and prompts.

The gap is that we persist the derived **cost**, not the **count** — the credits ledger stores `amount` + `metadata.model`, and `generated_assets.costMicros` likewise. Dividing back out by `unitPrice` works until fal moves a price, then silently corrupts.

Steps:

1. Persist `unitsBilled` alongside cost. All three call sites already hold it.
2. Have `update-fal-pricing.ts` prefer our observed median units, falling back to fal's historical estimate, then to nothing.
3. Make `estimateFalCost` return an explicit "unknown" rather than a fabricated default when it has no count for a `compute_seconds` model, and log loudly so the next such model surfaces instead of silently reading $0.003.
4. Re-introduce cost in the "Add a model" UI as an opt-in display, omitting the figure for models with no observations yet.

## Notes

- Step 3 is the honesty fix and stands alone — printing nothing beats printing a number that is 100× wrong.
- When pricing returns to the UI it needs a quality signal beside it. `qualityRank` already orders the list but is invisible, so an unopposed price column argues for the cheapest model with nothing arguing for the good one.
- Unrelated, spotted while investigating: `package.json` wants `@tanstack/ai-fal ^0.9.12`, `node_modules` has `0.9.10`.

---

## Amended plan (2026-07-29): DB-backed pricing store

Discussed while implementing: instead of mediating observed units through the regen script, pricing becomes a **live D1 store refreshed by the Worker cron**, provider-generic so LLM (OpenRouter) rows can join later.

**Schema** — two new platform-global tables (pure derived cache, fully rebuildable):

- `model_pricing` — snapshot keyed `(provider, endpointId, unit)` (unit in the key leaves room for LLM input/output token rates as two rows). Carries `unitPriceMicros` (verbatim from the provider), `typicalUnitsPerCall` (fal's historical estimate), `observedMedianUnits` + `observedSampleCount` (median `unitsBilled` from our own transactions — the preferred estimation signal), `fetchedAt`.
- `model_pricing_history` — appends a row whenever a unit price changes (and on first sight): a time-series of model costs.

**Refresh** — a second cron expression (`17 3 * * *`, daily) on the existing `scheduled()` handler runs `refreshFalPricing()`: fetches fal's pricing + historical APIs (shared client in `src/lib/ai/fal-pricing-fetch.ts`, also used by the regen script), computes 90-day median `unitsBilled` per endpoint from `transactions.metadata`, upserts the snapshot (chunked for D1's 100-param cap) and appends history rows on price change.

**Persistence (step 1 as planned)** — `falUsageMetadata()` spreads `endpointId` + `unitsBilled` into the transaction metadata of every fal-backed deduction (image ×10 sites, music, motion, upscale).

**Estimation (step 3 as planned, plus the live merge)** — `estimateFalCost` accepts a pricing map and prefers `observedMedianUnits` → `typicalUnitsPerCall`; for `compute_seconds` with neither it returns **null** (logged loudly) instead of the fabricated 3-second default. `getEffectiveFalPricing()` merges the D1 snapshot over the baked-in `FAL_PRICING` seed (5-min per-isolate cache); every credit-gate call site passes it and wraps with `gateEstimate()`, which substitutes a conservative $0.10/call floor when the estimate is unknown — over-gating slightly until the model's first observation lands, instead of under-gating ~100×.

**Deferred to a follow-up** — step 4 (re-introducing the opt-in price display in "Add a model"): per the note above it needs a quality signal designed next to the price column, and estimates should be allowed to self-correct from observations first.

tombeckenham:
Closes #1069

## Problem

Pre-flight cost estimates for fal models billed in compute seconds were up to ~98× low (Grok Imagine showed ~\$0.0026 for 5 scenes vs ~\$0.25 actual) because `estimateFalCost` fell back to a fabricated `DEFAULT_COMPUTE_SECONDS = 3` when fal's historical API returned no data — true for all six of our compute-seconds endpoints. This under-fed the pre-flight credit gate. **Billing was never affected** (actual charges use fal's reported `unitsBilled`).

## Approach (amended plan on the issue)

Pricing becomes a **live D1 store refreshed by the Worker cron**, self-correcting from our own observed usage:

- **`model_pricing` / `model_pricing_history`** — provider-generic tables keyed `(provider, endpointId, unit)` (room for OpenRouter LLM input/output token rates later, no migration needed). Snapshot rows carry the live unit price, fal's `typicalUnitsPerCall`, and our observed median `unitsBilled` + sample count; history appends whenever a unit price changes. Pure derived cache — fully rebuildable. Migration is additive `CREATE TABLE` only.
- **Daily refresh** — second cron expression (`17 3 * * *`) on the existing `scheduled()` handler runs `refreshFalPricing()`: fal pricing + historical APIs (client extracted to `src/lib/ai/fal-pricing-fetch.ts`, shared with the regen script), 90-day median `unitsBilled` per endpoint mined from `transactions.metadata`, chunked upserts under D1's 100-param cap.
- **Observation capture** — every fal-backed deduction (10 image workflows, music, motion, upscale) spreads `falUsageMetadata()` = `{ endpointId, unitsBilled }` into transaction metadata.
- **Honest unknowns** — `DEFAULT_COMPUTE_SECONDS` is gone. `estimateFalCost` prefers `observedMedianUnits` → `typicalUnitsPerCall` and returns **null** (logged loudly) when a compute-seconds model has neither. Gate sites pass `pricing: await getEffectiveFalPricing()` (D1 snapshot merged over the baked-in seed, 5-min per-isolate cache) and wrap with `gateEstimate()` — a conservative \$0.10/call floor, so unknown models over-gate slightly until their first observation lands instead of under-gating ~100×.

Estimators stay synchronous (optional pricing-map param) so pure callers and unit tests need no DB.

## Deferred (noted on the issue)

- Re-introducing the opt-in price display in "Add a model" — per the issue it needs a quality signal designed beside the price column, and estimates should self-correct from observations first.
- OpenRouter pricing rows in the same tables.

## Testing

- `bun tsgo --noEmit`, `bun lint`, `bun dead-code` clean.
- All 1952 unit tests pass, including new `estimateFalCost` cases (unknown compute-seconds → null; observed median beats historical; observed drives Grok to its real ~\$0.05/image).
- Migration applied to local D1.

---

## Review follow-ups (2 commits)

A multi-agent review of the original commit found the refresh pipeline could quietly re-create the bug it fixes — every failure mode wrote a null and continued. Two follow-up commits address the findings.

### `5e0c3eeb` — make the refresh fail loudly instead of degrading

- **Transient fal failure no longer destroys good pricing.** `fetchHistoricalCostUsd` returns a discriminated `'ok' | 'no-history' | 'failed'`; collapsing those into `null` meant one 429 nulled `typicalUnitsPerCall`, and — combined with a wholesale-replace merge in `getEffectiveFalPricing` — gated `gpt-image-2` at **$1.00/image instead of $0.22** (#1062) until the next successful refresh. The cron now preserves stored values for unreachable endpoints, aborts if >25% failed, and the live merge is field-wise over the seed.
- **`normalizeUnit` throws instead of defaulting to `'flat'`.** `unit` selects the entire estimation branch, so a guess is off by orders of magnitude — an unrecognised unit would estimate Grok Imagine at ~$0.00017 instead of ~$0.05, i.e. #1069 restored through the new DB path.
- **`MIN_OBSERVED_SAMPLES` before our median outranks fal's historical figure.** `observedSampleCount` was stored and never read, so a single aborted-but-billed run became the platform-wide gate. `EffectiveFalPricing.observed` now nests median + count so one can't be read without the other.
- **`gateEstimate` logs every floor substitution** with model and operation (deduped per isolate — it runs inside per-shot loops). Nothing could previously answer how often the gate ran on a made-up number.
- **`numImages` recorded and divided out** when mining medians: `unitsBilled` is per call, estimation is per image, and one call renders up to 4.

Also fixed: `verify-fal-costs.ts` rendered unknown estimates as `$0` — understating the total printed immediately before it spends real money — and reported microdollars in USD columns (1,000,000× off); `refresh-fal-pricing.ts` used literal NUL bytes as key separators, making the file **binary to git** (259 lines that never appeared in a diff); dropped a no-op `as number` that stripped the `Microdollars` brand; corrected `fal_model_pricing` → `model_pricing`, one instance of which the generator would have baked into a checked-in file.

### `6020696d` — bill from live prices, observe every generation

- **Billing reads the live table.** `falCostFromUnits` resolves through `getEffectiveFalPricing` (optional map so pure callers and tests skip D1). A fal price rise was previously detected, recorded in `model_pricing_history`, and still under-charged until someone re-ran the generator. The cron now warns on every change.
- **Observations moved to `model_usage_observations`**, written before every early return in `deductWorkflowCredits`. BYOK, zero-cost and insufficient-credit generations write no transaction, so mining the ledger only learned from teams in good standing without their own key — the models least likely to have an estimate were the least likely to gain one. No `teamId`: anonymous model telemetry, so no FK and clear of the D1 cascade trap.
- **Per-endpoint sample cap** via `ROW_NUMBER() OVER (PARTITION BY endpoint_id)`, replacing a global 5000-row scan that let one busy model starve every quiet one. Verified against the real D1 engine, not just the libsql harness — it's the first window function in the codebase.
- **Staleness is detectable**: `getEffectiveFalPricing` errors when the newest row is >48h old. A six-month-broken cron and a healthy one were previously indistinguishable.
- **Two guard tests**: cron-string presence in both `wrangler.jsonc` blocks (a drifted expression silently falls through to the reconcile sweep), and the D1 bound-param arithmetic asserted against real column counts — libsql has no 100-param cap, so it can't catch a breach (#1019).

Migration is additive only: one `CREATE TABLE` plus an index.

### Considered and rejected

Billing insufficient-credit teams and letting balances go negative was explored, but it required dropping the `positive_balance` CHECK — which SQLite can only do via a table rebuild of `credits`. Reverted; the constraint remains as the backstop under the application-level gates.

### Verification

`bun tsgo --noEmit` clean · **1960 tests / 191 files passing** · `bun lint` 0 errors · `bun dead-code` clean · migration applied to local D1 with no schema drift.


tombeckenham:
## Review round 3 — `5e564fc9`

A multi-agent review, plus **actually running** the new refresh against the live fal API, found this pipeline could not have worked in production.

### The blocker: fal rate-limits `/pricing/estimate` far harder than `/pricing`

Measured directly: even **strictly sequential** requests 800ms apart start returning 429 after the third.

At the shipped `ESTIMATE_CONCURRENCY = 5` / 150ms gap, **28 of 33 endpoints returned 429 on every run** — permanently over `MAX_TYPICAL_FETCH_FAILURE_RATIO`. The cron aborted before writing anything, so `model_pricing` would have stayed empty forever, indistinguishable from a healthy table while everything silently fell back to the seed.

Now one at a time, ~1s apart, with backoff honouring `Retry-After`:

| | rows | with `typicalUnitsPerCall` |
|---|---|---|
| before | 33 (only after the write-ordering fix below) | **4** |
| after | 33 | **26** |

The 7 without are the genuine no-history compute-seconds endpoints — exactly the population this PR describes.

### Billing could silently charge $0

`falCostFromUnits` returned `ZERO_MICROS` when the live-pricing read threw. Every call site guards deduction behind `cost > 0`, so `deductWorkflowCredits` — the only thing that reports a zero charge — was never reached. A D1 blip meant free generations platform-wide, with one log line.

It now bills from the static seed (exactly what an empty `model_pricing` already resolves to, so the failure path matches the cold-start path), and both remaining zero branches call `reportMissingBillingCost`. `motion-workflow` also priced *outside* any `step.do`, so a failed read couldn't even replay.

### Other paths that wrote a plausible snapshot and logged success

- **Empty price list wiped the table** via the stale sweep → throws first.
- **A broad historical-fetch failure discarded live prices AND our own medians** (which come from D1 and don't need fal at all) → writes both, then throws.
- **`warnIfStale` keyed off the NEWEST row**, so a nightly partial write reported healthy forever → keys off the oldest, names the stale endpoints.
- **`resp.json()` outside the try** → one HTML error page took down all 33.
- **Observations carry no `unit`**, so a re-denomination medians `compute_seconds` and `images` together into a *confident, non-null,* ~100× wrong number — #1069 through the mechanism built to prevent it → samples discarded when an endpoint's unit changes.
- **`getEffectiveFalPricing` resurrected a `typicalUnitsPerCall` the cron deliberately nulled**, contradicting the `ok`/`no-history`/`failed` split.
- **Duplicate endpoint rows picked a winner arbitrarily** — a wrong *charge* → dropped, so it gates on the floor and reports.
- **`music-workflow` dereferenced metadata its neighbours guard** → `TypeError` after a paid generation.
- **`recordFalUsage` swallowed its error**, so the enclosing `step.do` always succeeded and never used its retry.

### Also

`gateEstimate` emits `billing_estimate_floored` so floor usage is countable · `numImages` from `imageUrls.length` (a partial return biased the median low) · 13 copy-pasted observation blocks → `recordFalUsageStep` · `scripts/refresh-fal-pricing.ts` fills the table locally (`bun dev` never fires `scheduled()`).

### Tests

New coverage for the fal pricing client, refresh orchestration, and the `ROW_NUMBER` window function — **each mutation-verified**: swapping the `compute second`/`second` order fails, `DESC`→`ASC` fails, window→global `LIMIT` fails.

The audio cost assertions were tautological (`X - Y === X - Y`) — tripling the music cost left the whole suite green. `audioContribution` is now computed independently, and that mutation fails two tests.

### Considered and rejected

- `falCostFromUnits → Microdollars | null` — the seed fallback plus `reportMissingBillingCost` gets the observability without rippling `!== null` through 13 call-site guards. Worth doing as its own change.
- Branding `FalEndpointId` — the highest-value remaining type fix (it's the join key across four tables with zero enforcement, and `motion-workflow` already uses two different strings for one generation), but it touches DB rows, model configs and every workflow.
- Deleting the two "mixed selection" tests as redundant — once `audioContribution` became independent they catch the mutation, so removing them would lose coverage.

### Verification

`bun tsgo --noEmit` clean · `bun lint` **0 errors** · **1995 tests / 193 files** passing · `bun dead-code` clean · migration untouched and still additive-only.

Note: `src/lib/ai/fal-pricing-data.ts:2` still points at the old `UNITS_KIND` location. It's generated and the template is already correct, so it self-heals on the next `bun scripts/update-fal-pricing.ts` — hand-editing would be overwritten and would make a real price change unreviewable.

Issue #1067, opened 2026-07-25
Data model audit: remove redundant projections and mirror columns


tombeckenham:
## Problem

Several tables cache a copy of the row their selection pointer points at. The intent is a read-path convenience — draw the storyboard without a join. The cost is that every one of these creates two sources of truth, and a cache that starts being read for decisions is no longer a cache.

This is not hypothetical. It already produced a real bug (see the sibling issue): `frames.imageModel` is documented as "display/cache only", but the model resolver reads it to pick the generate model, so a turbo preview render silently pins full-quality regeneration to `flux_2_turbo`.

The dual-source design also already needs divergence handling — `src/lib/db/scoped/shot-prompt-versions.ts:306` logs `"points at motion prompt version … but no row exists (orphaned pointer); falling back to the cached mirror"`.

## Audit scope

Every table with a `selected_*_version_id` pointer plus mirrored columns. Known systems:

| # | Mirror | Source | Written by |
|---|---|---|---|
| 1 | `frames.imageUrl`, `imagePath`, `previewImageUrl`, `imageStatus`, `imageGeneratedAt`, `imageError`, `imageModel`, `imageInputHash` | `frame_variants` | `buildFrameImageMirror` (`scoped/frames.ts:52`) |
| 2 | `shots.videoUrl`, `videoPath`, `videoStatus`, `videoGeneratedAt`, `videoError` | `video_variants` | `buildShotVideoMirror` (`scoped/shots.ts:39`) |
| 3 | `shots.motionPrompt`, `motionPromptInputHash` | `shot_prompt_versions` | `mirrorOntoShot` (`scoped/shot-prompt-versions.ts:131`) |
| 4 | `frames.imagePrompt`, `visualPromptInputHash` | `frame_prompt_versions` | `framePromptVersions.select` |
| 5 | `sequences.musicPrompt`, `musicTags`, `music*` | `sequence_music_prompt_versions`, `sequence_music_variants` | `scoped/sequence-music-prompt-versions.ts:115` |

Plus a related case that isn't even a mirror: **`frames.imageWorkflowRunId`**. `frame_variants.workflowRunId` already exists (`schema/frame-variants.ts:65`), and each variant has its own generation run, so the frame column is a lossy last-writer-wins collapse — 38 of 82 local frame/variant pairs already disagree. Workflow run id is per-variant by nature and should not exist on the frame.

## Evidence the caches aren't earning their keep

Measured on local D1 for the `frames` case:

[...]

Issue #990, opened 2026-06-25
SSF redesign Phase 3: video_variants + 15s render segments


tombeckenham:
Part of the **Scene → Shot → Frame redesign**. Full design: `docs/architecture/scene-shot-frame-redesign.md`.

**Phase 3 — Video variants + 15s segments.** Blocked by Phase 1 (and Phase 2 for the primary-frame anchor).

## Scope
- `video_variants` — flat versions keyed `(sceneId, model, shotSetKey)`, carrying an ordered **manifest** snapshotting exactly what the render consumed (per-shot `motionPromptVersionId`, `frameVersionId?`, `durationMs`). `inputHash` over the manifest → O(1) staleness.
- Render unit ≠ scene: a scene's video is an ordered tiling of **≤15s segments**, each a contiguous shot-subset (model cap is 15s). Scene ≤15s ⇒ one segment; long scenes split; per-shot render = degenerate segment.
- `scene.renderPlan: [{ shotIds, selectedVideoVersionId }]` — revert-a-segment = repoint. (Explicit `render_segments` entity is the upgrade path; default `renderPlan`.)
- Per-shot video mirrored on the shot (`shot.videoUrl` + `selectedVideoVersionId`); staleness re-routed through the primary frame.
- **`shot_variants` retires** (video → `video_variants`; audio deferred — only sequence music exists for now).

## Folds in / supersedes
- **#910** — stage 4: multi-shot scene rendering on capable models.
- **#953** — activate multi-shot scene rendering: per-shot image pipeline + render activation.

## Acceptance
- Typecheck + motion / segment tests.

Issue #612, opened 2026-04-29
Drizzle-kit d1-http migrator cascades data loss on table rebuilds


tombeckenham:
## Summary

The standard SQLite "table rebuild" pattern that drizzle-kit emits (`PRAGMA foreign_keys=OFF` → `CREATE __new_X` → `INSERT SELECT` → `DROP X` → `RENAME`) silently cascade-deletes data when applied to Cloudflare D1.

We hit this in production on 2026-04-29 with migration `20260428013041_productive_kabuki` (PR #607, drizzle ORM v1 upgrade). Every `ON DELETE CASCADE` FK pointing at `user.id` cascade-fired:
- `team_members` wiped
- `session` wiped (only handful of post-migration logins survived)
- `account` wiped (only post-migration logins survived)
- `passkey`, `team_invitations` already empty

`user`, `teams`, `sequences`, `frames` survived (no FK or no CASCADE).

Recovery via slug-correlation INSERT (all users restored from intact `teams` table). Backup at `backups/velro-prd-20260429-143920.sql`.

## Root cause (verified, not hypothesized)

Verified by reading installed drizzle-kit `1.0.0-beta.22` source and Cloudflare workers-sdk issue [#5438](https://github.com/cloudflare/workers-sdk/issues/5438) (closed 2026-04-15 with explanation from Cloudflare engineer alsuren):

1. **drizzle-kit's d1-http migrator joins all migration statements with `"; "` and sends them in ONE HTTP request** to D1's `/query` endpoint (verified in `node_modules/drizzle-kit/bin.cjs:234549-234553`).
2. **D1's `/query` endpoint auto-wraps the multi-statement body in an implicit transaction** (per Cloudflare engineer comment).
3. **SQLite silently ignores `PRAGMA foreign_keys = OFF` inside a transaction** — per [SQLite docs](https://sqlite.org/foreignkeys.html): _"It is not possible to enable or disable foreign key constraints in the middle of a multi-statement transaction. Attempting to do so does not return an error; it simply has no effect."_
4. **`PRAGMA defer_foreign_keys = ON` does not prevent CASCADE actions** — only defers constraint checks. Per [Cloudflare docs](https://developers.cloudflare.com/d1/sql-api/sql-statements/#pragma-defer_foreign_keys--onoff): _"setting PRAGMA defer_foreign_keys = ON does not prevent ON DELETE CASCADE actions from being executed."_
[...]

PR #1105, opened 2026-08-05
feat(#1067): phase 2a — backfill render_segments + selection so shots.video* stops being the source of truth

tombeckenham:
Phase 2a of the #1067 data-model audit. Part of #1067 — this PR **drops no column**; the drop of `shots.video*` is a separate PR, after this is deployed and re-measured on prod.

## The problem

The #990 cutover left the video surface half-migrated on purpose. Its own header says:

> the segment's selection pointer stays NULL for legacy rows and the next render/select repoints it

It also only created `render_segments` for shots that had a `shot_variants` row with `variant_type='video'`, so shots whose video predates that table got no segment at all.

Users don't re-render old work. Measured read-only on `openstory-prd` 2026-08-05:

| | count |
|---|---|
| shots with `video_url` | **3041** |
| …with **no** `render_segment_id` | **1041** |
| …with a segment whose `selected_video_version_id` is NULL | **1887** |
| …fully wired | **113** |
| `render_segments` total / with a selection | 2074 / **113** |

**`shots.video*` is the primary record for 2928 of 3041 videos — not a mirror.** Re-routing reads through the segment pointer before this backfill would blank them. That's what makes the eventual column drop unsafe today, and what this PR fixes.

## What it does

Four guarded steps, `INSERT`/`UPDATE` only:

1. Degenerate one-shot `render_segments` row for each of the 1041 orphan shots
2. `video_variants` row built from the mirror columns, manifest matching #990's shape
3. Point the segment at its variant; link the shot to the segment
4. For the 1887 legacy segments, set `selected_video_version_id` to the variant whose url its shot serves today

## Preconditions verified on prod before writing the SQL

Exact counts, not estimates:

- Of the 1041 orphans: **0** have a NULL `motion_model` (satisfies `video_variants.model NOT NULL`), **0** have a NULL `scene_id` (satisfies `render_segments.scene_id NOT NULL`), all 1041 urls distinct, all have a non-null `video_path` and `video_generated_at`.
- Of the 1887 segments needing a pointer: **all 1887** resolve to **exactly one** non-discarded `video_variants` row by url match — 0 unmatched, 0 ambiguous.

## Judgement call, 14 rows

1027 of the 1041 orphans are `video_status: 'completed'`; **14 are `'failed'` while still carrying a playable url, path and `generated_at`** — a later attempt failed without clearing the asset.

All 1041 are written as `'completed'`. The variant row records *the asset that exists*; the failed attempt gets no variant row because its data was never retained beyond `video_status`/`video_error`. Writing those 14 as `'failed'` would make their segment's selected version fail any completed-check and blank a video that plays today. Documented in the migration header and covered by a test.

## Safety

- **Set-based `UPDATE … FROM`, not correlated subqueries.** The first draft used per-row `EXISTS`/scalar subqueries; `check-migrations.ts` rejected them. A per-row subquery over `render_segments` (2074) / `shots` (4329) trips D1's remote CPU-time limit (7429), and because migrations apply *before* deploy, that freezes the whole deploy (#1019). Rewritten so the driver is scanned once and the target is searched by primary key. The guard now reports clean.
- **`INSERT`/`UPDATE` only** — no table rebuild, no `DROP`, so no inbound `ON DELETE CASCADE` fires (#612).
- **Replay-safe** — segment and variant ids reuse the shot id (the #990 precedent for segments), and every statement is guarded with `WHERE NOT EXISTS` / `IS NULL`, so a re-run is a no-op. There's a test for it.
- **No fabricated hashes** — `input_hash` stays NULL; every one of these rows has a NULL `video_input_hash` on prod, and inventing one would read as "freshly verified".

## Tests

8 tests in `src/lib/db/backfill-video-segments.test.ts`, executing the shipped SQL **verbatim** from the migration file against an in-memory DB (the same pattern as the existing #990 backfill test, but reading by explicit path rather than scanning — a scan would pick whichever migration sorted first now that two of them insert into `video_variants`):

- manifest key shape — the #1 risk, since keys must match `VideoManifestEntry` or every reader breaks on data that passed SQL
- the 14-row failed-status case
- discarded variants ignored even when the url matches
- segments that already have a selection left untouched
- replay idempotency
- a coverage assertion that no shot with a video is left without a resolvable selection

Full suite: **200 files, 2064 tests passed**. Typecheck 0 errors, lint 0 errors, format clean.

## How this gets verified against real data before merge

Not "merge and hope". The rehearsal is a full dress run on a copy of prod:

1. `wrangler d1 export openstory-prd --env=production --remote --output=prod.sql` (read-only)
2. Load that dump into local D1 and apply the migration there
3. Assert the coverage condition on the copy: shots with a `video_url` whose selection doesn't resolve to a variant with that url must be **0**, from a starting point of 2928
4. Confirm the 1041/1887 split moved exactly as measured, and that no pre-existing selection changed

Because every statement is additive and guarded, the failure modes are bounded: it can add a wrong row, never delete or overwrite one. The only column it writes on an existing row is `render_segments.selected_video_version_id` **where it is currently NULL**, and `shots.render_segment_id` where it is currently NULL. Nothing that has a value today is modified.

Rollback, if it comes to that, doesn't need a database restore — the rows it creates are identifiable (`render_segments.id = shots.id`, `video_variants.id = shots.id`) and the pointers it sets were NULL, so a targeted undo is exact. D1 Time Travel is the backstop rather than the plan, since restoring the whole DB would discard unrelated writes made since.

## Follow-up before the drop

Deploy this, then re-measure on prod. `count(*)` of shots with a `video_url` whose selection doesn't resolve must read **0** before phase 2d re-routes reads and drops the 8 columns (`video_url`, `video_path`, `video_status`, `video_workflow_run_id`, `video_generated_at`, `video_error`, `video_input_hash`, `motion_model`).

Phase 2d also removes the `buildShotVideoMirror` write that stamps each shot's `durationMs` with the segment's **summed** duration while `sequences.ts:317` and `motion-functions.ts:127` read it as per-shot.


tombeckenham:
## Rehearsed against a full copy of production

`wrangler d1 export` of `openstory-prd` (94M) loaded into a scratch SQLite DB, migration applied there. The copy reproduced the prod baseline exactly before anything ran — 4329 scenes / 4329 shots / 4329 frames, 2074 `render_segments`, 2100 `video_variants`.

### Coverage

| metric | before | after |
|---|---|---|
| shots with `video_url` | 3041 | 3041 |
| …with **no** `render_segment_id` | 1041 | **0** |
| `render_segments` with a selection | 113 | **3041** |
| **shots whose video does not resolve through the pointer** | **2928** | **0** |
| `render_segments` total | 2074 | 3115 (+1041) |
| `video_variants` total | 2100 | 3141 (+1041) |

Exactly 1041 rows added to each table — the orphan count, no more, no less.

### Nothing pre-existing was disturbed

- **0** of the 113 pre-existing selections changed value (snapshot diffed before/after).
- **0** foreign-key violations across the entire database afterwards (`PRAGMA foreign_key_check`).
- **0** variants with invalid JSON in `manifest`; **0** missing the `shotId` key that `VideoManifestEntry` requires.
- **0** NULL `model` values (the NOT NULL target).
- All 1041 backfilled variants written as `completed`, including the 14 whose shot carries `video_status: 'failed'` with a playable asset.

### Idempotency, on real data

Applied a **second** time: `render_segments` 3115 → 3115, `video_variants` 3141 → 3141, unresolvable 0 → 0. Exact no-op, matching the unit test.

### Runtime

0.18s wall / 0.03s user CPU for the whole backfill. That's the number the set-based rewrite was for — the original correlated-subquery version is what `check-migrations.ts` rejected, since a per-row subquery over these tables trips D1's remote CPU limit (7429) and freezes the deploy, because migrations apply *before* it (#1019).

Issue #1087, opened 2026-07-31
Carry error codes across the server-fn boundary; drop message-sniffing


tombeckenham:
## Problem

Server functions throw typed errors (`OpenStoryError` subclasses carrying a `code`), but only `error.message` survives the server-fn boundary — `code`, `name`, `statusCode` and `details` are all lost. The client arrives at a plain `Error` and has to reconstruct intent by sniffing prose.

This is already documented in two places as a known workaround:

- `src/components/models/model-detail-view.tsx:69-74` — *"Server-fn errors arrive as plain `Error`s (CatalogApiError's status/code don't survive serialization)"*
- `src/lib/errors.ts` — the `isInsufficientCreditsError` fallback

It has already caused one real bug. `isInsufficientCreditsError` matched on `error.message.includes('Insufficient credits')`, which also matches OpenRouter's own message — pinned in `src/lib/ai/llm-client.test.ts:744`:

```
Insufficient credits. Add more using https://openrouter.ai/settings/credits
```

That's the *provider's* balance under a BYOK key. The false positive opened our billing dialog, sent the user to top up an account that was already fine, and discarded the real error message so the actual problem was unreportable.

#1081 patched this by stamping a `[INSUFFICIENT_CREDITS]` marker into the message and matching that. It works and it's server-controlled, but it's a workaround with a real cost: an internal token now lives inside a user-facing string, and every site that renders a raw `error.message` has to strip it via `errorMessage()`. Five files that can receive a credits error still display raw messages and would leak the token:

- `src/components/scenes/scene-music-facet.tsx`
- `src/components/scenes/shot-duration-field.tsx`
- `src/components/element/element-selector.tsx`
- `src/components/element/element-token-button.tsx`
- `src/components/script/script-view.tsx`

## Proposed fix

Carry the code across the boundary properly, then delete message-sniffing entirely.

Server functions throw a `Response` with a JSON body containing the structured error (`OpenStoryError.toJSON()` already produces exactly the right shape — `{ name, message, code, statusCode, details }`). The client unwraps it once in a shared helper and gets a typed object with a real `code` field.

Rough shape:

```ts
// server
throw json({ error: err.toJSON() }, { status: err.statusCode });

// client
const code = errorCode(error);            // reads the parsed body
if (code === 'INSUFFICIENT_CREDITS') { … }
```

## Scope

- 8 `OpenStoryError` subclasses (`src/lib/errors.ts:37-99`)
[...]

Issue #1085, opened 2026-07-29
Update all: durable stale-artifact regeneration + pre-created pending artifact records (in-flight staleness, dedup, cancellation)


tombeckenham:
> **Scope update (2026-07-31).** This issue now covers two phases. The
> **"Update all" feature itself** was split out of #1077/#1081 and lives here
> (Phase 0, implemented on branch
> `1085-pending-artifact-records-in-flight-staleness`, stacked on #1077).
> The pre-created-pending-records design below is Phase 1 — it hardens what
> Phase 0 ships. #1077/#1081 is now scoped to the staleness UI only.

## Phase 0 — "Update all" (implemented, awaiting review)

Durable server-side regeneration of every artifact in scope that reads stale
right now. Branch: `1085-pending-artifact-records-in-flight-staleness`.

- `UpdateStaleShotsWorkflow` — shot / scene / sequence scope. One artifact per
  stale flag, **no cascade**: regenerating a visual prompt outdates the shot's
  image, but that image is only re-rendered if it was already stale itself.
  Video never auto-renders.
- The plan (which shots, which artifacts) is computed from live state in a
  `compute-plan` step and frozen as its durable result. It holds **ids and
  flags only** — scene bodies are materialised per shot in `prepare-prompt-*`,
  because inlining them blew Cloudflare's 1 MiB step-result cap on a
  sequence-scope run where every shot is stale.
- Credit safety: `regenImage` requires an existing rendered image, so "Update
  all" never spends credits creating a *first* still. Credits are preflighted
  before enqueue — the last point at which the client can still be shown a
  billing dialog — and the in-workflow failure is `NonRetryableError`.
- Failures are reported, not logged into the void: `getWorkflowRunOutcome`
  reads a run's terminal status **and its output**, and the client polls it
  rather than inferring completion from indicators clearing.
- Staleness gained an `'unknown'` state (the comparison itself failed),
  distinct from `'untracked'` (no hash on file). A write path must not treat
  "we couldn't tell" as "nothing to do", so those shots are reported as
  `skipped`.

### Known gaps carried into Phase 1

- **No dedup.** Nothing stops a second click, a second tab, or a teammate
  enqueuing duplicate work — the guard is client-local. This is exactly what
  Phase 1's pending records fix.
- **`prepare-image` re-reads current state** after the prompt child lands
  (last-write-wins), so a post-click edit can leak into the run. Phase 1
  makes a run a pure function of its click-time plan.
[...]

Issue #827, opened 2026-06-05
talent/location detail-page e2e specs fail locally on a clean checkout


tombeckenham:
While verifying #771 I ran the full e2e replay suite on an UNMODIFIED tree (main-equivalent baseline, r2-mock still in place) and got 9-10 consistent failures, all in the same cluster:

- `locations.spec.ts` › Edit Location › can view location detail page / open edit dialog / cancel edit dialog
- `locations.spec.ts` › Location Library - List View › can navigate between location detail pages
- `talent.spec.ts` › Edit Talent with Reference Media › (view detail / open edit / add media / displays media / cancel)
- `talent.spec.ts` › Talent with Media - List View › can navigate between talent detail pages

Symptoms:
- `getByRole('heading', { name: 'Reference Images' })` not found on the detail page, or
- the test never leaves the talent/locations LIST page (click on the card doesn't navigate), so the edit button is never found.

These same specs presumably pass in CI (suite is green on main), so this looks like a local-environment difference (macOS, `workers: 4`, local `colorScheme: 'dark'`?) rather than a product bug. Filing separately so #771 stays focused — confirmed NOT caused by the #771 changes (identical failures with the r2-mock sidecar still present).

Repro: `bun test:e2e:setup && bun test:e2e` on a clean checkout, macOS.

PR #1107, opened 2026-08-05
1067: drop the last shots/scenes mirrors + snapshot workflow inputs at the trigger (WorkflowScopedDb)

tombeckenham:
Part of #1067 (milestone 19 data-model audit — more phases to follow, so no closing keyword).

Two bodies of work on this branch:

# 1. Drop the last shots/scenes mirrors + delete the ShotWithImage projection

Continues the audit: every remaining mirror column on `shots` and `scenes` is gone, and the compat projection that flattened joined rows into invented names is deleted. Each drop was verified against production data first.

### Prompt mirrors — `frames.image_prompt` / `visual_prompt_input_hash`

These could not simply be deleted because `input_hash` was doing two jobs: recording the upstream context a prompt aligns with (what staleness compares) **and** acting as the dedupe slot for a UNIQUE index on `(frame_id, input_hash)`.

Retry idempotency is really `(frame, input_hash, text)` — same context *and* same output. So a force-regen (same context, new text) collided with the index and had to insert with `input_hash NULL` to escape it, throwing the alignment fact away. The mirror existed to carry it back.

`write` now matches all three columns before inserting, so every row keeps its real hash and the mirror has nothing left to carry. The UNIQUE indexes become plain lookup indexes; the live-claim UNIQUE index is untouched, since that is what guards concurrent double-spend (#1085).

This also fixes a latent regression in the earlier motion-side drop: with `shots.motion_prompt_input_hash` already gone, a force-regenerated motion prompt resolved to a null hash and would have read stale forever.

### `shots.order_index`

An exact copy of `scenes.order_index`, reached through the shot's own scene link — the split writes both from the same loop variable. Verified in production: all 4334 shots satisfied `shots.order_index = scenes.order_index`, `shot_number = 1`, zero null `scene_id`, zero duplicate `(scene_id, shot_number)`.

It also carried the upsert conflict target that makes a workflow replay hit the same shot row (`dedup-ids.ts`). That moves to `(scene_id, shot_number)`, which the split already knows at insert time. The new unique index is **partial** so neither column needs `NOT NULL` — that would force a `shots` table rebuild, and `frames` / `shot_prompt_versions` / `shot_variants` all cascade off it (#612).

Most display sites needed no replacement: they already read `scene ? scene.orderIndex + 1 : shot.orderIndex + 1`, so the drop deleted a dead fallback.

### `scenes.original_script`

Duplicated the content of the version row `selected_script_version_id` points at. All 4334 production scenes have a selected version, so the fallback was dead — and **3 had already diverged**, rendering stale script text. Removing the mirror fixes them.

The trap: `seedSplitFromSceneRows` seeded the version rows *by reading the column*, so a naive drop would have silently stopped scene splitting from creating versions at all, giving every new scene an empty script.

### `scenes.music_design`

Dropped — write-dead since March. The only writer is `buildSceneInsert`, called only from scene-split, but `sceneSplittingResultSchema` picks just sceneId/sceneNumber/originalScript/metadata plus an extended `continuity`. `musicDesign` is not in the split LLM's output, so the insert always wrote NULL. `motion-music-prompts-workflow` does compute music design, but attaches it to an in-memory scene that is never persisted.

Production confirms it: all 657 rows with a value were created between 2026-03-13 and 2026-03-25; every one of the 3677 scenes created since is NULL.

An earlier commit on this branch kept the column and taught the canonical summary builder to read it, on the evidence that it *was* read and that 657 rows held data. That was the wrong test — the right question was whether a second mechanism had taken over the job, and the split schema had quietly stopped emitting the field. Reverted here, along with the music prompt paragraph instructing the model to reason about `musicPresence`, which nothing supplied.

What survives independently: there were **two builders producing different summaries for the same prompt**, and the duplicate also wrote `sceneId: shot.id` — the shot's id in the scene's field. That stays deleted.

The in-memory `Scene['musicDesign']` field is untouched — analyze-script reads it live off the motion-music workflow's result to decide whether to generate music. Only the column goes.

### `ShotWithImage` -> `ShotView`

The projection flattened joined rows into a third vocabulary — `frame.imageStatus` became `thumbnailStatus`, `frameVariant.url` became `thumbnailUrl`. `ShotView` exposes the rows as themselves. Six projected fields had no production reader at all and are gone. `videoStatus` is kept, because it is genuinely *derived* (newest primary render wins, selection is the fallback) rather than renamed.

The v1 API needed no change — it already served a nested `{ image, video }` shape.

**One place the layer was doing real work:** `query-cache-updater.ts` wrote urls straight onto the cached shot from realtime events, inventing a URL with no row behind it. A `ShotView` cannot fabricate a `frame_variants` row, so handlers now patch the url only when the row is cached and refetch to land the real one, reusing the existing debounce key. Frame-owned fields still patch live, so status stays instant.

### Bugs fixed in passing

- `getShotsFn` resolved a frameless shot's motion prompt then dropped it; the sibling reads in sequences/admin both passed it.
- A re-split producing fewer scenes detached the surplus shots and deleted only the scenes, leaving rows every read fetched and no view rendered. `order_index` concealed this by giving a dead row a position. They are deleted now.
- `sequences.getWithShots` had no production callers — only a test asserting it delegates.
- `failure-summary-banner` rendered `Scene {f.orderIndex + 1}` off the *shot* index; the field is now `sceneNumber`, from the scene.
- The duplicate music builder set `sceneId: shot.id` — the shot's id in the scene's field.
- Storybook fixtures carried per-shot prompts but passed a null prompt source to the projection, so every prompt was silently discarded before reaching the UI.

### Known follow-ups (not in this PR)

- `scripts/generate-scenes-view-fixture.ts` predates the row split: it selects the dropped `shots.order_index` and builds no frame/variant rows, so running it would fail *and* overwrite the hand-maintained fixture. The fixture header now says so instead of claiming to be auto-generated.
- A re-analyze never appends a new `split` script version (`seedSplitVersions` skips scenes that already have one), so a re-analyzed script is not shown. Predates this work; the mirror merely made it look stored.
- `SceneThumbnail` / `VariantSelector` / `VideoStateOverlay` keep props named `thumbnailUrl` etc. Those are presentational component APIs, not ShotView reads — renaming them is a separate, wider sweep.

### Migrations

All additive or `ALTER TABLE ... DROP COLUMN` / index ops — no table rebuilds, so the #612 cascade trap does not apply. Data was backfilled into version rows before each mirror drop.

### Gates

typecheck, 2061 unit tests, lint (0 errors), knip, and migrations applying cleanly from an empty D1 with no schema drift.

https://claude.ai/code/session_01P3feZ9wy3DHNcnThRCs5e5

# 2. Snapshot workflow inputs at the trigger; enforce via WorkflowScopedDb

### Summary

A 16-agent audit found ~30 distinct critical mid-run mutable-D1 reads across 16 Cloudflare Workflows — the rule ("content-generation workflows must not read mutable DB state for anything frozen at trigger time") existed in `docs/architecture/workflow-snapshots-and-content-hash-staleness.md` but nothing enforced it. This PR fixes every finding and makes the class unrepresentable going forward.

### Fixes (highlights)

- **Billing correctness**: LLM/fal key resolution moved inside the step that spends it (`llm-call-helper` + hand-rolled copies) — `usedOwnKey` is now replay-stable, so a BYOK key added/removed mid-run can no longer invert the charge decision across six workflows.
- **Motion manifest**: recorded the pre-write selection pointer on every user-edit run (deterministic bug, not a race) — now records the version this run actually wrote/pinned.
- **Storyboard/analyze**: the whole `AnalyzeScriptWorkflowInput` was manufactured from a mid-run read while `triggerStoryboard` discarded the row it already held — payload now built at the launcher; element set pinned by trigger-time `elementIds`.
- **Recast**: regenerate snapshots built at the trigger server fns; a diverged sheet now skips regeneration and reports `sheetDiverged` instead of regenerating against the old identity and returning hardcoded success.
- **Update-all**: `computePlan` moved to the trigger; the plan rides the payload as pure ids/flags/hashes (~205 KB at 100 shots; prompt text, standing still/motion selections, and `durationMs` all pinned at click time and read back by explicit id).
- **Upscale/image**: `promptVersionId` stamped from the trigger so still↔prompt pairing survives upscales; `shot-images` renders and hashes from the same element set.
- **Export API**: stitch list + music resolved in the POST handler — "still generating" is now a synchronous 4xx instead of an async failed export.
- **Element vision**: token snapshotted at the trigger with a compare-and-swap cascade, so a mid-flight user rename wins instead of being silently overwritten.
- Legacy payload fallbacks replaced with static defaults; trigger-side throws are typed `OpenStoryError`s (ride the #1087 adapter after rebase onto main).

### Enforcement (prevention by construction)

- **`WorkflowScopedDb`** (`src/lib/db/scoped-workflow.ts`): `runImpl` receives the write surface with all read methods removed — a mid-run read is a compile error. Sanctioned reads go through three named hatches whose spelling is the justification: `scopedDb.credentials` (secrets), `scopedDb.claims` (append-only rows by explicit id — cannot express a selection pointer), `scopedDb.liveRead` (genuinely live by design: divergence recomputation, polls, existence guards, billing guards, balance).
- **`src/lib/workflow/no-mid-run-reads.test.ts`**: per-file allow-list of every surviving read, bucketed and justified; fails on new reads, on stale entries, and on a read arriving through the wrong hatch. Compile-time assertions keep the type itself honest.
- CLAUDE.md + the architecture doc now state the rule and point at the enforcement.

### Gates

- `bun typecheck` clean, `bun lint` 0 errors, `bun format:check` clean, knip clean
- `bun run test`: **208 files, 2200 tests, 0 failures**
- e2e: 33 passed; the 11 failures are pre-existing issue #827 (verified identical with all changes stashed — no regressions from this work)

Follow-up filed: #1113 (unify sheet/element divergence onto the #1085 pending-claim pattern; closes the hash-recompute blind spots).

https://claude.ai/code/session_01P3feZ9wy3DHNcnThRCs5e5


tombeckenham:
## Summary

A 16-agent audit found ~30 distinct critical mid-run mutable-D1 reads across 16 Cloudflare Workflows — the rule ("content-generation workflows must not read mutable DB state for anything frozen at trigger time") existed in `docs/architecture/workflow-snapshots-and-content-hash-staleness.md` but nothing enforced it. This PR fixes every finding and makes the class unrepresentable going forward.

## Fixes (highlights)

- **Billing correctness**: LLM/fal key resolution moved inside the step that spends it (`llm-call-helper` + hand-rolled copies) — `usedOwnKey` is now replay-stable, so a BYOK key added/removed mid-run can no longer invert the charge decision across six workflows.
- **Motion manifest**: recorded the pre-write selection pointer on every user-edit run (deterministic bug, not a race) — now records the version this run actually wrote/pinned.
- **Storyboard/analyze**: the whole `AnalyzeScriptWorkflowInput` was manufactured from a mid-run read while `triggerStoryboard` discarded the row it already held — payload now built at the launcher; element set pinned by trigger-time `elementIds`.
- **Recast**: regenerate snapshots built at the trigger server fns; a diverged sheet now skips regeneration and reports `sheetDiverged` instead of regenerating against the old identity and returning hardcoded success.
- **Update-all**: `computePlan` moved to the trigger; the plan rides the payload as pure ids/flags/hashes (~205 KB at 100 shots; prompt text, standing still/motion selections, and `durationMs` all pinned at click time and read back by explicit id).
- **Upscale/image**: `promptVersionId` stamped from the trigger so still↔prompt pairing survives upscales; `shot-images` renders and hashes from the same element set.
- **Export API**: stitch list + music resolved in the POST handler — "still generating" is now a synchronous 4xx instead of an async failed export.
- **Element vision**: token snapshotted at the trigger with a compare-and-swap cascade, so a mid-flight user rename wins instead of being silently overwritten.
- Legacy payload fallbacks replaced with static defaults; trigger-side throws are typed `OpenStoryError`s (ride the #1087 adapter after rebase onto main).

## Enforcement (prevention by construction)

- **`WorkflowScopedDb`** (`src/lib/db/scoped-workflow.ts`): `runImpl` receives the write surface with all read methods removed — a mid-run read is a compile error. Sanctioned reads go through three named hatches whose spelling is the justification: `scopedDb.credentials` (secrets), `scopedDb.claims` (append-only rows by explicit id — cannot express a selection pointer), `scopedDb.liveRead` (genuinely live by design: divergence recomputation, polls, existence guards, billing guards, balance).
- **`src/lib/workflow/no-mid-run-reads.test.ts`**: per-file allow-list of every surviving read, bucketed and justified; fails on new reads, on stale entries, and on a read arriving through the wrong hatch. Compile-time assertions keep the type itself honest.
- CLAUDE.md + the architecture doc now state the rule and point at the enforcement.

## Gates

- `bun typecheck` clean, `bun lint` 0 errors, `bun format:check` clean, knip clean
- `bun run test`: **208 files, 2200 tests, 0 failures**
- e2e: 33 passed; the 11 failures are pre-existing issue #827 (verified identical with all changes stashed — no regressions from this work)

Follow-up filed: #1113 (unify sheet/element divergence onto the #1085 pending-claim pattern; closes the hash-recompute blind spots).

Part of #1067.

https://claude.ai/code/session_01P3feZ9wy3DHNcnThRCs5e5

PR #1162, opened 2026-08-12
feat: Seedance 2.5 and Seedream 5 Pro on fal (BytePlus via stays off)

### LINKED ISSUE #1157: Add byteplus natively
Use the new TanStack AI adapter for BytePlus (`@tanstack/ai-byteplus`) to call Seedance (video) and Seedream (image) **directly against BytePlus Ark** instead of proxying them through fal. Add to the model catalog and to sequence creation. Handle Ark's rate limits.

## Why

Seedance and Seedream are ByteDance models. Today we reach both through fal (`bytedance/seedance-2.0/enterprise/v2/*`, `fal-ai/bytedance/seedream/v5/lite/*`), which means fal's margin, fal's queue, and fal's parameter subset. Calling Ark directly gives us BytePlus billing and quota, the first-class Seedance request fields (`camera_fixed`, `generate_audio`, `return_last_frame`, `service_tier`, reference-media roles), and BytePlus's own model ids.

## Decisions (agreed 2026-08-12)

1. **Routing — repoint the existing keys, keep fal as fallback.** Sequences store model *keys* (`seedance_v2`), not endpoint ids, so the route behind a key can change without touching existing rows. Add a `via: 'fal' | 'byteplus'` discriminator to the model config. Prefer BytePlus when `ARK_API_KEY` is configured; fall back to the existing fal endpoint otherwise. No duplicate picker entries, no breakage for fal-only deployments.
2. **Credentials — platform key only.** `ARK_API_KEY` as a Worker env var / secret. **No** BYOK for BytePlus in this issue: `team_api_keys` stays `'openrouter' | 'fal'`. A team's own fal key still routes their Seedance through fal (which is the correct behaviour — they're paying fal).
3. **Pricing — static rate card seeded into `model_pricing`.** The adapter exposes no pricing API, and Seedance bills in tokens (`usage.total_tokens` off the finished task) rather than fal's billable units. Hardcode BytePlus's published per-model rates, seed them into `model_pricing` under a new `'byteplus'` provider, and charge `reported units x rate`. No new cron; the existing estimate/credit-gate path keeps working.
4. **Rate limits — per-provider fan-out cap + 429-aware backoff.** Ark enforces per-account RPM and concurrency quotas. Split `FANOUT_CONCURRENCY` so BytePlus work fans out at its own (lower, env-tunable) cap, and add 429/quota-error retry with exponential backoff that does **not** consume the content-flag retry budget. `service_tier: 'flex'` (half price, offline queue, Seedance 1.x only) is explicitly out of scope for now.

## Scope

**In:** Seedance video + Seedream image via Ark. Model catalog, sequence-creation pickers, generation paths, pricing rows, rate limiting, tests.

**Out:** Seed chat models (LLM stays on OpenRouter), Seed Speech TTS/ASR, BYOK for BytePlus, `service_tier: 'flex'`, a BytePlus billing-reconcile cron.

### Model catalog
- [ ] Add `@tanstack/ai-byteplus` dependency.
- [ ] Add `via: 'fal' | 'byteplus'` to `IMAGE_TO_VIDEO_MODELS` / `IMAGE_MODELS` entries in `src/lib/ai/models.ts`, with the BytePlus model id alongside the fal endpoint id so a key can resolve either way.
- [ ] Pick the BytePlus ids: Seedance `dreamina-seedance-2-0-260128` (matches today's fal 2.0 enterprise route) and Seedream `seedream-5-0-260128`. `dreamina-seedance-2-5-260628` is real but 404s with `ModelNotOpen` until activated in the Ark console — gate it behind activation rather than shipping it dark.
- [ ] Resolver that picks the route per call: BytePlus when `ARK_API_KEY` is set and the model has a BytePlus id, else fal.

### Generation paths
- [ ] `src/lib/motion/motion-generation.ts` — branch `submitMotionJob` / `pollMotionJob` on `via`, using `byteplusVideo(...)` for the Ark route. Map our `aspectRatio` + `duration` onto Ark's `ratio_resolution` size template and its per-model duration range (the adapter's `snapDuration` replaces our `MOTION_JSON_SCHEMAS` snap for BytePlus models).
- [ ] Reference images: Ark takes `reference` / `character` prompt-part roles rather than fal's tagged `image_urls[]` with `@ImageN` tokens. Build the BytePlus content array in a sibling of `build-model-input.ts` — the fal codegen (`bun motion:codegen`) does not and will not cover Ark.
- [ ] `src/lib/image/image-generation.ts` + `build-image-request.ts` — BytePlus branch. Seedream takes size tokens (`2K`) or explicit pixels, has no `n` (multi-image is `sequential_image_generation` group mode, an upper bound not a count), carries references inline rather than on a separate edit endpoint, and **defaults `watermark` to true** — pass `watermark: false` explicitly.
- [ ] Generated video/image URLs expire after 24h; confirm the existing R2 download path runs well inside that window.

### Billing
- [ ] Widen `ModelPricingProvider` to include `'byteplus'` (text column, no migration).
- [ ] Rate-card module with BytePlus's published Seedance/Seedream rates, sourced from the BytePlus pricing docs at implementation time and dated in a comment. Seed into `model_pricing`.
- [ ] Cost path: exact charge from the task's reported usage x rate; record `model_usage_observations` rows so the observed median works the same way it does for fal.
- [ ] `/pricing` catalog rows gain a `BytePlus` **via** value.

### Rate limits
- [ ] Per-provider fan-out caps in `src/lib/workflow/map-in-waves.ts` (env-tunable), applied wherever image/motion children fan out.
- [ ] 429 / Ark quota-error classification + exponential backoff retry, distinct from the content-flag re-roll path in `motion-workflow.ts`.

### Config, tests, docs
- [ ] `ARK_API_KEY` in `.env.example`, env typing, and `bun setup`.
- [ ] E2E: aimock intercepts fal via the `x-fal-target-host` header, which Ark requests don't carry. Point the adapters' `baseURL` at aimock under `E2E_TEST` and record fixtures for the Ark endpoints.
- [ ] Unit tests: route resolution, BytePlus request building (size/duration/reference roles), pricing math, 429 backoff.
- [ ] Update `CLAUDE.md` — the fal-integration section currently reads as if fal is the only media provider.

## Risks

- **Region isolation.** Ark keys are region-scoped and Seedance is served only from `ap-southeast`; an EU key silently fails. Worth a startup assertion.
- **Aggressive model retirement.** BytePlus retires dated model ids fast. The fal fallback is the safety net when an id goes dark.
- **Rate-card drift.** Static rates can go stale in either direction. Mitigation is a dated comment and a follow-up issue for reconcile-against-invoice if drift shows up.

### LINKED ISSUE #1216: Extract a shared media-provider seam; stack the BytePlus/OpenAI/Grok PRs on it
Three provider PRs (#1162 BytePlus, #1170 OpenAI, #1171 Grok) each hand-roll the same machinery into the shared media files: an if-branch in `submitMotionJob`/`pollMotionJob`/`image-generation.ts`, their own key-resolution rule, URL prep, billing normalization, and a provider stamp on the submission (`route: MediaRoute` vs `provider: MotionProvider` — same concept, two names). They all collide on the same files, and the if-ladder grows per provider.

## Proposal

Land a behavior-preserving base PR off `main` that extracts a provider seam, then rebase the three PRs into thin "add one provider module" changes.

All routes already converge on TanStack AI `generateVideo`/`getVideoJobStatus`, so the seam is thin — a `MotionProvider` (+ image sibling):

- `claim(modelKey, scopedDb) → ResolvedApiKey | undefined` — each route's own rule; ordered registry, first claim wins, fal is the always-claiming fallback
- `submit(options, key) → { jobId }` / `poll(jobId, key)` — provider owns request building, URL prep, and retry policy internally
- billing hooks: `pricingEndpointId(modelKey)`, `costFromUsage(usage)`
- orchestrator stamps one `provider` field on the submission; poll dispatches on the stamp (poll must follow the submission — job ids are provider-scoped). Retire `route`.

Base PR also absorbs the shared-file churn: the submission `provider` field, poll dispatch, extensible `team_api_keys` provider enum.

Derive the interface from the three branch diffs, not speculatively. Provider-specific complexity (BytePlus quota backoff, tokens÷1000, pricing aliasing; xAI exact cost) stays inside each module — the seam gives it a place, not an eraser. Text-LLM side is out of scope: `create-adapter.ts` is already that seam.

## Sequencing

1. Base PR off `main` — seam + fal as the only provider (pure refactor)
2. Rebase #1171 (Grok) — smallest, early-return shape maps 1:1 onto `claim()`
3. Rebase #1170 (OpenAI) — exercises the image side
4. Rebase #1162 (BytePlus) — biggest, but its provider files are already module-shaped

tombeckenham:
**BytePlus Ark is not being rolled out.** `ARK_API_KEY` is unset in CI and production. Seedance and Seedream keep going through fal. The native via is in the tree and stays off until we set that key on purpose.

Related to #1157. Does **not** close #1216.

## What this PR actually ships (fal)

- Seedance **2.5** on the existing `seedance_v2` catalog key (`bytedance/seedance-2.5/image-to-video` + `reference-to-video`). Sequences store the key, not the endpoint.
- Seedream **5.0 Pro** on `seedream_v5` (`bytedance/seedream/v5/pro/{text-to-image,edit}`).
- Studio fal siblings, 2.5 pricing seed, lazy-load of the Ark adapter so preview deploys don’t CPU-timeout at startup.
- Official Seedance 2.5 prompt skill (`sd25-pe`).

## What’s in the code but off

A `via: 'fal' | 'byteplus'` path on the same catalog keys (Grok native-via pattern). Claim order: native model + live `ARK_API_KEY` + team is **not** on its own fal key. BYOK fal stays on fal. Poll follows the job stamp; missing stamps default to `'fal'`.

Also present, also inert without Ark/IAM keys: studio Ark submit/poll, `asset://` still ingest (CreateAsset), quota backoff, static BytePlus rate card.

Turning it on later is setting `ARK_API_KEY` (and IAM keys if we want `asset://`). Do **not** do that in CI or prod as part of this merge. Entry Advanced Creation Rights is 50 unique stills; paid capacity is a separate SKU.

## Why we’re not turning it on

Ark’s photoreal-face filter needs `asset://` ingest. That quota is per BytePlus **account**, not per team, and paid Advanced Creation Rights is expensive. fal remains the production via.

## E2E

Stays on fal. `isBytePlusConfigured()` is false under `E2E_TEST` unless `ARK_BASE_URL` names a mock.

## Not in this PR

Seed chat, Seed Speech, BytePlus BYOK, `service_tier: 'flex'`, a BytePlus billing-reconcile cron, enabling Ark in prod/CI.


tombeckenham:
Rebase onto the #1222 via/vendor seam is in progress (agent). Target: widen `MediaVia` with `'byteplus'`, retire `route` in favor of stamped `via`, catalog `vendor` + `byteplusId`, BYOK fal stays on fal.

Issue #839, opened 2026-06-06
Look into workflow failures


tombeckenham:
Yesterday we had 240 workflow failures. Concerneed the durable object changes have broken something. Look at posthog data to see if we can unravel it, and look at surfacing errors and retries to users

tombeckenham:
## Investigation findings (PostHog logs + CF instance data, 2026-06-07)

**TL;DR: the Durable Object realtime change (#802) is exonerated.** The ~240 failures were a cascade triggered by a single API user bulk-creating ~50 sequences in 17 seconds, amplified by a Cloudflare Workflows engine restart and three weaknesses in our fan-out architecture.

### Timeline (June 6, UTC / AEST)

| Time (UTC) | Time (AEST) | Event |
|---|---|---|
| 04:36:50–04:37:07 | 14:36–14:37 | One user (`01KTD0RZ5B…`, team `01KTD0RZ7R…`, created the same day) bulk-creates **~50 sequences in 17s** — UGC-ad-style scripts, consistent with the new public one-shot API (#795, live since June 5, rate limit 10/s). Each storyboard spawns analyze-script + ~8 child workflows → **~400–500 concurrent instances** + hundreds of concurrent LLM/image calls. |
| 04:38:16 | 14:38 | PR #832 preview deploy lands (`pr-832` worker + secrets push). |
| 04:41:35–04:42:09 | 14:41–14:42 | **Workflows engines mass-abort**: 244 log lines `Failure: Aborting engine: Grace period complete` (`cf.outcome: canceled`) across StoryboardWorkflow + children. No prod deploy occurred at this time (prod deployments on June 6: 07:33 and 11:31 UTC only) — this was either a CF-platform engine restart (possibly load-triggered) or a side-effect of the preview deploy; every other grace-abort cluster on June 5–6 also coincides with a deploy window. |
| 04:42–05:07 | 14:42–15:07 | Instances resume (CF replays from step cache); about half the storyboards complete. |
| 05:07:12–05:07:37 | 15:07 | ~20 storyboard parents hit the **30-minute `waitForEvent` timeout** (`WorkflowTimeoutError: Execution timed out after 1800000ms`) waiting for their analyze-script child → errored. `StoryboardWorkflow.onFailure` is **log-only**, so these sequences are stuck in `processing` forever — no user-visible failure, no retry affordance. |
[...]

PR #1261, opened 2026-08-24
fix(#1259): region fallback for Anthropic-blocked users, DeepSeek default for China

### LINKED ISSUE #1259: Storyboard pipeline hard-fails for users in Anthropic-blocked regions (no model fallback)
## What happened

A new signup (China, Zhejiang; residential IP; Chrome/Windows) signed up at 2026-08-24 10:19 UTC and submitted the same 1,678-char script **9 times in 3 minutes**. Every storyboard run died server-side with:

```
LLM stream error [model=anthropic/claude-opus-5-fast]: This model is not available in your region.
```

Cascade per run: `SceneSplitWorkflow` → `AnalyzeScriptWorkflow` → `StoryboardWorkflow` all failed. `recommendStylesForScriptFn` failed the same way on `claude-sonnet-5`; `claude-opus-5` hit it once too. The user saw nothing useful in the UI and just kept clicking create — the Slack "Film / sequence created" spam was retries, not abuse.

The server-side LLM call egresses from the Cloudflare colo nearest the user (likely HKG here), so Anthropic's geo-block applies even though the call is "server-side". DeepSeek calls in the same session succeeded, and image gen (hunyuan/flux via fal) ran fine — only the Anthropic-model steps die.

**Not a one-off:** the same error appears 10 more times across the previous 7 days from other sessions (`openstory` prod Worker logs, `$metadata.error includes "not available in your region"`). Anyone in mainland China/HK etc. currently cannot generate a storyboard when the pipeline picks an Anthropic model.

## Fixes

1. **Fallback:** on a region-unavailable error, the LLM client should retry with a region-available model (DeepSeek is already in the model mix) or a provider route that isn't geo-blocked. Right now this error class exhausts step retries and kills the run.
2. **Surface the failure:** the composer gave no visible error, which is what produced 9 identical submissions. Show the failure state.
3. **Debounce/dedup creation:** `triggerWorkflow` already supports `deduplicationId` — hash (teamId, script, settings) over a short window so retry-clicking can't spawn N identical pipelines. Also disable the create button while the mutation is pending.

## Evidence

- Worker logs (`openstory`, 2026-08-24 10:15–10:40 UTC): ~30 region errors across ≥4 workflow runs, e.g. run ids `01M0SMPT7YPHW7SB2YXDR0BSBM`, `01M0SMPNBSKSPMCWEVH4F9MYQK`, `01M0SMSRHENXHMDAG4DH0NR8R3`, `01M0SMPGSWXNW67K658QSKDR98`.
- PostHog Production: 9 `sequence_generated` events 10:22:02–10:24:45 UTC, identical `script_length=1678`, auto-style/motion/music all true; ~$0.22 LLM spend; no motion generation ever started.
- PostHog `$exception` captured none of this — client-side error tracking is blind to workflow failures.

tombeckenham:
Closes #1259.

## What

A signup from mainland China had every storyboard run die on \`This model is not available in your region\` — our LLM calls egress from the Cloudflare colo nearest the user, so Anthropic's geo-block applies even to "server-side" calls. No fallback existed, the composer showed nothing, and the user resubmitted the identical script 9 times.

## Changes

**Region policy (new \`src/lib/ai/region-policy.ts\`)**
- Request-time: \`resolveModelForCountry\` swaps any \`anthropic/*\` model to DeepSeek when the request's \`cf-ipcountry\` is a known Anthropic-blocked country (CN/HK/MO/RU/BY/IR/KP/SY/CU). Applied in \`createSequences\`, so a China-region sequence never persists an Anthropic analysis model.
- Error-time: any LLM call that still hits a region block retries once on a region-available model — DeepSeek for text, Mistral Small 4 for vision-bearing calls. Wired into \`callLLMStream\` (guarded so a retry can never replay already-yielded content) and both durable workflow helpers (\`withRegionFallback\`), so steps recover instead of exhausting retries. This also covers the hardcoded \`SCENE_SPLIT_MODEL\` boundary pass, which has no request context to swap proactively.

**DeepSeek bumped to latest**
- \`deepseek/deepseek-v3.2\` → \`deepseek/deepseek-v4-pro-0813\` (1M context, structured outputs, newest on OpenRouter as of 2026-08-24). v3.2 stays in the registry \`hidden\` for existing sequences; pricing data regenerated via \`bun scripts/update-openrouter-pricing.ts\`.

**Composer feedback + dedup**
- \`useCreateSequence\` now toasts on failure (it was fully silent).
- \`createSequences\` rejects a resubmit when every requested analysis model already has a live (<2 min, \`processing\`) run of the exact same script for the team — retry-clicking can no longer spawn N identical pipelines. (The Generate button was already disabled while pending; per-sequence runs were already mutexed by #839 — the gap was N fresh sequences.)

## Notes / deliberate scope cuts

- Anthropic can still be *attempted* once from a blocked region on paths with no request context (workflow-internal \`SCENE_SPLIT_MODEL\`, \`RECOMMENDED_MODELS\` server fns) — the attempt fails at Anthropic's edge and immediately falls back, so no user-visible failure and no Anthropic output is ever served there. Threading country through workflow payloads wasn't worth the churn.
- PostHog \`$exception\` blindness to workflow failures (issue evidence #3) not addressed here.

## Tests

- New \`region-policy.test.ts\` (error matching, fallback picks, country swap, retry semantics).
- \`llm-client.test.ts\`: fallback retry uses DeepSeek adapter; mid-stream region error does not retry.
- Full unit suite (2,685), typecheck, lint, knip all green.

### DIFF
--- a/src/components/model/model-selector.tsx
+++ b/src/components/model/model-selector.tsx
@@ -1,11 +1,17 @@
 import { BaseModelSelector } from './base-model-selector';
+import { getRequestCountryFn } from '@/functions/ai';
 import {
   isSelectableAnalysisModelId,
   isValidAnalysisModelId,
   SCRIPT_ANALYSIS_MODELS,
   type AnalysisModelId,
 } from '@/lib/ai/models.config';
-import { useMemo } from 'react';
+import {
+  isRegionBlockedModel,
+  resolveModelForCountry,
+} from '@/lib/ai/region-policy';
+import { useQuery } from '@tanstack/react-query';
+import { useEffect, useMemo } from 'react';
 
 const GROUP_ORDER = ['all'] as const;
 
@@ -22,10 +28,21 @@ export const ModelSelector: React.FC<ModelSelectorProps> = ({
   disabled = false,
   singleSelect = false,
 }) => {
+  // Region-blocked vendors are hidden from the picker (#1259): a user in an
+  // Anthropic-blocked country shouldn't be offered a model the server would
+  // swap away anyway. While the country is loading, the unfiltered list shows
+  // — the server-side swap in createSequences stays the enforcement.
+  const { data: country } = useQuery({
+    queryKey: ['request-country'],
+    queryFn: () => getRequestCountryFn(),
+    staleTime: Infinity,
+  });
+
   const models = useMemo(
     () =>
       [...SCRIPT_ANALYSIS_MODELS]
         .filter((m) => isSelectableAnalysisModelId(m.id))
+        .filter((m) => !isRegionBlockedModel(m.id, country))
         .sort((a, b) => a.qualityRank - b.qualityRank)
         .map((m) => ({
           id: m.id,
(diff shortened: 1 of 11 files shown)


tombeckenham:
Added in 9d00d3e2: the analysis-model picker now hides region-blocked (Anthropic) models — `getRequestCountryFn` exposes the request's `cf-ipcountry`, `ModelSelector` filters the list and remaps any already-selected blocked model (stored default / restored draft) onto the DeepSeek fallback. Server-side `resolveModelForCountry` remains the enforcement; the picker filter is purely so users are never offered a model the server would swap away.

PR #1264, opened 2026-08-24
Remove Grok and xAI integrations

visoar:
## Summary
- remove Grok text, image, and video models plus native xAI routing
- remove xAI key management, dependencies, generated schemas, fixtures, and documentation
- add a migration that deletes stored xAI team API keys

## Validation
- 278 test files / 2721 tests passed
- TypeScript typecheck passed
- oxlint passed with existing warnings only
- knip passed
- migration safety check passed
- production build passed


visoar:
Superseded by UllrAI/openstory#11, which rebased the removal onto the project main after the Prism, credits, dependency, default-model, and localization merges, passed the full CI suite, and was merged as 2ab145b3.

Issue #1167, opened 2026-08-12
Use grok natively


tombeckenham:
Use grok natively through tanstack-ai instead of using openrouter and fal.

`@tanstack/ai-grok` (the xAI adapter) covers every Grok surface we currently rent from another provider:

| Surface | Today | Native |
| --- | --- | --- |
| Chat/LLM | `x-ai/grok-4.5`, `x-ai/grok-4.20` via `@tanstack/ai-openrouter` | `createGrokText` → xAI Responses API (`grok-4.5`, `grok-4.20-0309-reasoning`) |
| Image | fal `xai/grok-imagine-image/quality/{text-to-image,edit}` | `createGrokImage` (`grok-imagine-image-quality`) — generation + image-conditioned edits |
| Video | fal `xai/grok-imagine-video/v1.5/image-to-video` | `createGrokVideo` (`grok-imagine-video-1.5`) — async job + poll, image-to-video only |

## Decisions

- **Scope:** all three surfaces go native — chat, image, and video.
- **Keys:** new platform secret `XAI_API_KEY`, **plus BYOK** — `'xai'` joins `API_KEY_PROVIDERS` so a team can bring its own xAI key through the existing `team_api_keys` / `resolveKey` path, with the same validate / mark-invalid / platform-fallback semantics as OpenRouter and fal.
- **Fallback:** with no xAI key resolvable (fresh clone, deploy-button, PR previews), Grok models keep routing through OpenRouter/fal exactly as they do today. Native is opt-in on key presence, so `bun install && bun dev` still works on a fal-only key.
- **Billing:** trust adapter-reported usage/cost, with xAI's published rates filling the gaps the adapter leaves.

## Billing detail (the non-obvious part)

Native reporting is uneven, so each surface bills differently:

- **Video** — the adapter returns exact cost (`usage.cost_in_usd_ticks`, i.e. $0.08/second for `grok-imagine-video-1.5`). Use it directly.
- **Chat** — the adapter reports token counts only, no cost. OpenRouter's per-request `usage.cost` is what `llmCostFromUsage` bills on today, so the native path loses it and must price tokens against xAI's published rates ($2/1M in, $6/1M out for grok-4.5; the 200k-token tier bump is real and needs handling or explicit acceptance).
- **Image** — the adapter reports token counts only, no cost, and `GROK_IMAGE_MODELS` carries no pricing. Bill at xAI's published per-image rate ($0.05/image for `grok-imagine-image-quality`).

[...]

Issue #895, opened 2026-06-11
Promote byok


tombeckenham:
I'd like to make it possible to use just a fal key and route openrouter calls through fal's openrouter endpoint. I'd also like to change the dialog and byok setup so that fal shows a fal logo and allows the user to enter the key direct in the dialog. I'd also like it to show connect with open router and an openrouter logo direct in that dialog too. Add buy credits under those two, and gift token down the bottom

PR #1266, opened 2026-08-24
feat(#1265): route LLM calls through the LLMTR gateway

### LINKED ISSUE #1265: Add LLMTR as a BYOK LLM provider (Turkey-hosted, OpenAI-compatible gateway)
## Description

**Disclosure up front: I'm the owner of llmtr.com.** Take the framing below as an interested party's, and the technical claims as things you can verify against the endpoints named.

### What

[LLMTR](https://llmtr.com) is a Turkey-hosted, OpenAI-compatible LLM gateway. One key fronts Anthropic / OpenAI / Google / xAI / DeepSeek / Z.ai / Mistral, plus models it hosts in Turkey itself. I'd like it available as a BYOK provider alongside OpenRouter, fal and xAI.

### Why it's cheap to add

It speaks OpenRouter's wire format **and** serves an OpenRouter-shaped `/v1/models` (`pricing.prompt`, `architecture`, `top_provider`, `supported_parameters`). So the existing `@tanstack/ai-openrouter` adapter drives it with nothing but a `serverURL` swap — exactly the trick #895 already plays with fal's OpenRouter proxy. No new SDK, no new adapter.

Verified against the live gateway: streaming, `stream_options.include_usage`, and `response_format: json_schema` structured output all behave the same as OpenRouter's.

### Why it's worth adding

- **Data residency.** Requests terminate in Turkey. For TR/EU users with KVKK/GDPR constraints, "which jurisdiction did my script pass through" is currently unanswerable in OpenStory — every LLM path exits through OpenRouter or fal.
- **Region coverage.** Complements #1259: an alternative egress is useful wherever a first-party API is geo-blocked.
- **One more BYOK on-ramp.** Same argument as #1167 and #1157 — more ways for a team to bring a key they already have.

### Proposed scope (text/LLM only)

1. `llmtr` in `API_KEY_PROVIDERS` — encrypted team key + platform `LLMTR_API_KEY`. No migration needed (`provider` is `text`, no CHECK constraint).
2. `src/lib/ai/llmtr.ts`: registry id → LLMTR slug map, and a rate table.
   - **Slug drift is the trap.** LLMTR namespaces some vendors differently (`xai/` not `x-ai/`, `zai/` not `z-ai/`, `mistral/` not `mistralai/`). 14 of the 17 registry models map; the other three (`claude-opus-5-fast`, `deepseek-v3.2`, `seed-2.0-mini`) have no equivalent and should stay on OpenRouter/fal rather than being silently swapped for a near neighbour.
   - **Unaudited spend is the other trap.** LLMTR reports token counts but no per-request `cost`, so `llmCostFromUsage` needs the resolved `via` to price the call — otherwise every LLMTR call bills $0. Same shape and same caveat as the native Grok rates (#1167): it bypasses `model_pricing` and the hourly fal reconcile, so #1069 drift detection won't cover it.
3. Resolution order in `resolveLlmKey`: native xAI for Grok → team LLMTR key (when LLMTR carries the model) → team OpenRouter → team fal → platform.
4. Settings → API Keys row, and `LLMTR_API_KEY` in `.env.example` / `push-secrets.ts`.

Media (image/video/audio) stays on fal and xAI. LLMTR carries those endpoints too, but motion job ids are via-scoped (`MotionJobSubmission.via`), which makes it a separate surface — not part of this.

### Out of scope here

Putting LLMTR's Turkey-hosted models (`llmtr/gemma-4`, `llmtr/qwen3-6-35b`) in the model picker. Registry ids are assumed routable for everyone, so a sequence saved with an LLMTR-only model would become unrunnable the moment the key is removed. That needs an availability gate in the picker (a `requiresProvider` field or similar) and is worth its own issue.

I have this working and will open a PR against it.

knowhycodata:
## Related Issue

Closes #1265

## Summary of Changes

Adds [LLMTR](https://llmtr.com) — a Turkey-hosted, OpenAI-compatible LLM gateway — as a BYOK provider alongside OpenRouter, fal and xAI.

**Disclosure: I'm the owner of llmtr.com.** Everything technical below is verifiable against the live endpoints; the judgement calls are flagged so you can overrule them.

It speaks OpenRouter's wire format and serves an OpenRouter-shaped `/v1/models`, so the existing `@tanstack/ai-openrouter` adapter drives it with nothing but a `serverURL` swap — the same trick #895 plays with fal's OpenRouter proxy. No new SDK, no new adapter, no migration (`provider` is `text` with no CHECK constraint).

`src/lib/ai/llmtr.ts` exists to pin down the two things that would otherwise break silently:

**Slug drift.** LLMTR namespaces some vendors differently (`xai/` not `x-ai/`, `zai/` not `z-ai/`, `mistral/` not `mistralai/`), so every routable model is spelled out in `LLMTR_TEXT_MODELS`. 14 of the 17 registry models map. The other three — `claude-opus-5-fast`, `deepseek-v3.2`, `seed-2.0-mini` — have no LLMTR equivalent, and I deliberately did **not** alias them to a near neighbour: that would silently run a model the caller didn't ask for. They stay on OpenRouter/fal, which means an LLMTR-only team falls through to the platform key for those three. Flagging it as a judgement call in case you'd rather it fail loudly instead.

The four renamed slugs widen the OpenRouter model union via `extendAdapter`, the same mechanism `CATALOG_LAG_MODELS` uses. A renamed id necessarily loses that catalog's per-model metadata — including `OPENROUTER_COMBINED_TOOLS_AND_SCHEMA_MODELS`, so a call combining tools *and* structured output takes the adapter's fallback path on those four. The 11 identity mappings are unaffected.

**Unaudited spend.** LLMTR reports token counts but no per-request `cost`, so `llmCostFromUsage` now takes the resolved `via` and prices from `LLMTR_TEXT_RATES` — without it every LLMTR call would bill $0. Same shape and same caveat as the native Grok rates from #1167: it bypasses `model_pricing` and the hourly fal reconcile, so #1069 drift detection covers none of it. Rates were read from `llmtr.com/v1/models` on 2026-08-24 and currently match `OPENROUTER_PRICING` for every shared model (LLMTR takes its margin on credit top-ups, not per-model).

**Resolution order** (`resolveLlmKey`): native xAI for Grok → team LLMTR key *when LLMTR carries the model* → team OpenRouter → team fal → platform. A team that adds an LLMTR key chose that gateway deliberately, so it outranks their OpenRouter key. Key resolution and `createAdapter` both ask `llmtrTextModel`, so they can't disagree about routability.

**`validateKey` can't use `/v1/models`** — it's public and answers 200 for a bogus key, so it proves nothing. Validation is a 1-token completion on a $0 model; only 401/403 condemns the key.

Also: `fallbackReason` assignments in `resolveLlmKey` became `??=` so the reason reported is the highest-priority skipped key rather than the last one checked (behaviour-identical when no LLMTR key is present), and `resolveLlmKey` now passes `model` to `getPlatformLlmKey`.

Media (image/video/audio) stays on fal and xAI. LLMTR carries those endpoints, but motion job ids are via-scoped (`MotionJobSubmission.via`), so that's a separate surface.

## Risk Assessment

- [ ] Low
- [x] Medium
- [ ] High

It touches `resolveLlmKey` and `llmCostFromUsage`, which sit on every LLM call. Every new branch is gated on a provider key that no existing team or deployment has, so the behaviour for anyone not using LLMTR is unchanged — but the hot path is the hot path, hence Medium rather than Low. The `??=` change is the one edit with any reach, and it's behaviour-identical without an LLMTR key.

## Additional Notes

**Verified live against the gateway** (total spend $0.0032):

| Check | Result |
| --- | --- |
| Streaming text via `callLLMStream` | ✅ |
| Structured output (`response_format: json_schema`) | ✅ — works despite the catalog not advertising `structured_outputs` |
| All 4 renamed slugs end-to-end | ✅ `mistral-small-latest`, `zai/glm-5.2`, `xai/grok-4.6`, `xai/grok-4.20-0309-reasoning` |
| Billing from token counts | ✅ non-zero, matches catalog rates |
| `validateKey`: real key / bogus key | ✅ 200 / 401 |

**Test coverage:** 26 new unit tests across `llmtr.test.ts` (mapping + rates, including a guard that every mapped model has a rate and that `LLMTR_ONLY_MODEL_IDS` matches exactly the renamed slugs), `create-adapter.test.ts` (routing, slug translation, aimock precedence, platform-key order), `api-keys.test.ts` (resolution order, invalid-key skips, fallback reasons) and `llm-client.test.ts` (cost). `bun run test` ✅ · `bun typecheck` ✅ · `bun lint` ✅ · `bun run build` ✅ · `bun dead-code` ✅.

Two repo checks could not run on my machine (Node 20.18.1) and fail identically on a clean `main`, so they're not regressions: `no-bare-select-with-join.test.ts` needs `node:fs` `globSync` (Node ≥22), and `bun dead-code` needs `--bun` to get past knip's oxc-parser ESM require (it passes that way). CI should cover both.

**e2e:** never sets `LLMTR_API_KEY`, so recorded fixtures keep exercising the OpenRouter/fal path unchanged — no fixture refresh needed.

**Deliberately out of scope:** LLMTR's Turkey-hosted models (`llmtr/gemma-4`, `llmtr/qwen3-6-35b`) are not in the model picker. Registry ids are assumed routable for everyone, so a sequence saved with an LLMTR-only model would become unrunnable the moment the key is removed. That needs an availability gate in the picker and deserves its own issue.

## Screenshots (if applicable)

The LLMTR row in Settings → API Keys reuses the existing `ManualKeyRow`, rendering identically to the fal.ai and xAI rows (brand mark, status badge, key input, "Get a key from LLMTR" link).


tombeckenham:
Quick question on the slug map — could LLMTR accept OpenRouter ids as **aliases** of the canonical catalog ids? If `/v1/chat/completions` treated the left column as the right column, we could drop `LLMTR_TEXT_MODELS` / `extendAdapter` and drive the gateway with a `serverURL` + key the way we already do for fal's OpenRouter proxy.

I went and checked who actually owns these strings. **Neither catalog is using the first-party API ids for the vendor prefix.** Native APIs are unprefixed:

| First-party API | OpenRouter (our registry) | LLMTR catalog (today) |
| --- | --- | --- |
| `grok-4.6` ([xAI docs](https://docs.x.ai/docs/models)) | `x-ai/grok-4.6` | `xai/grok-4.6` |
| `grok-4.20-0309-reasoning` | `x-ai/grok-4.20` (collapsed) | `xai/grok-4.20-0309-reasoning` |
| `glm-5.2` / now `glm-5.3` ([Z.ai](https://docs.z.ai/guides/overview/quick-start) uses `glm-5.3` unprefixed) | `z-ai/glm-5.2` | `zai/glm-5.2` |
| `mistral-small-latest` ([Mistral](https://docs.mistral.ai/getting-started/models/models_overview)) | `mistralai/mistral-small-2603` (dated snapshot) | `mistral/mistral-small-latest` |

So: OpenRouter's `x-ai` / `z-ai` / `mistralai` are *its* author slugs, not xAI/Z.ai/Mistral's. LLMTR's `xai` / `zai` / `mistral` are *your* author slugs. For the bit after the slash, LLMTR is often closer to the provider (Mistral's `-latest` alias; xAI's actual `grok-4.20-0309-reasoning` build). We are not asking you to abandon those — only to also answer the OpenRouter-shaped ids our registry already speaks.

Aliases that would let us send registry ids unchanged:

| Accept (OpenRouter / our registry) | Route to (your catalog) |
| --- | --- |
| `x-ai/grok-4.6` | `xai/grok-4.6` |
| `x-ai/grok-4.20` | `xai/grok-4.20-0309-reasoning` |
| `z-ai/glm-5.2` | `zai/glm-5.2` |
| `mistralai/mistral-small-2603` | `mistral/mistral-small-latest` |

**Opus 5 Fast** is a different kind of gap. Anthropic has no `claude-opus-5-fast` model id — Fast is `model: "claude-opus-5"` + `speed: "fast"` ([docs](https://platform.claude.com/docs/en/build-with-claude/fast-mode)). OpenRouter packaged that as a separate slug (`anthropic/claude-opus-5-fast`) because it has no `speed` param. That slug is **not** on [llmtr.com/v1/models](https://llmtr.com/v1/models) today, and it is the model we always use for scene-split (`SCENE_SPLIT_MODEL`). Aliasing it to standard Opus 5 would be the silent neighbour-swap you correctly refused. If LLMTR can accept `anthropic/claude-opus-5-fast` and actually run Fast mode (2× price, ~2.5× OTPS), scene-split can stay on a team LLMTR key; otherwise we keep resolving that one call against OpenRouter/fal.

Happy to keep the client map if aliases are a no. Just wanted to ask before we bake a second catalog into the repo.

tombeckenham:
Hey, @knowhycodata, thanks so much for raising this PR. I don't mind this idea. I'm looking at making integrations like this simpler. Firstly, is LLMTR an OpenAI compatible interface? I'm also a maintainer at TanStack and we have an OpenAI compatible adapter. Would LLMTR work with that? See https://tanstack.com/ai/latest/docs/adapters/openai-compatible

knowhycodata:
Thanks for the pointer. Disclosure up front: I work on LLMTR.

Yes to both — LLMTR exposes a standard OpenAI-compatible interface, and that adapter should drive it directly.

The main blocker right now is version skew: @tanstack/ai-openai peers on @tanstack/ai@^0.49.1, while this repo is still on 0.47.2 and the three adapters currently installed here still peer on ^0.47.x. If useful, I can prepare a separate PR for that as well — bump the @tanstack/ai family first, then layer the adapter change on top.

On the slug map, I can remove the mechanical vendor-prefix differences from our side (x-ai/ → xai/, z-ai/ → zai/, mistralai/ → mistral/) by accepting them as aliases at the gateway. That would let OpenRouter-shaped IDs resolve without maintaining a per-model translation table.

We'd still need a small coverage list, since the catalogs don't fully overlap. For example, we don't currently carry claude-opus-5-fast, which is your SCENE_SPLIT_MODEL, so resolution still needs to know which IDs should go to LLMTR and which should remain on OpenRouter/fal.

We could also open a separate PR for Tencent models and add the Tencent models available through LLMTR using the same integration path.

And if you'd like to test the integration yourself, sign up at llmtr.com and I can add a small amount of trial credit to your account. Just send me the email you used, here or privately. email: data@knowhy.co or discord: toandto

tombeckenham:
I'll update the prod deps. There's a big pr with that

tombeckenham:
I will get this to this, I swear!

tombeckenham (approved):
I'm going to approve this. It works. Great to have it in there. It uses the openai compatible adapter.

Issue #1140, opened 2026-08-10
Transparent pricing


tombeckenham:
I want to differentiate openstory from other tools on the market that have "unlimited" that aren't so unlimited, and hidden charges. I want OOpenStory to be as open as possible with what it costs to create a film. Firstly, I want the option to include costs under any action. This should be on by default. So generate should show what it's going to likely cost. Then repeat for all other actions. There should be a toggle somewhere near where the wallet is turned on and off to disable this feature. I also want to indicate to the user that they get welcome credits when signing up. When signed out, show the wallet with $10 in it and highlight green. Then I want the pricing page to have a section on what it really costs to make a film.

Issue #1099, opened 2026-08-03
Billing gate improvement


tombeckenham:
The billing gate does not seem to work at all. The user gets an insufficient credit warning.  Also I want some design changes.

I want the gate to duplicate exactly how openai's API topup works. They've got the design down perfectly. Praise to the ux gods at OpenAI.

The billing gate should be a dialog, and it should be around the same size as the adding credits dialog. Then from the add credits button, open another dialog.

Adding credits should be a modal dialog, and it should look exactly like this.

<img width="479" height="376" alt="Image" src="https://github.com/user-attachments/assets/5c2ab702-ced9-47ad-b702-9188d8cb3042" />

A dialog should be the only way to add credits. Replace the current panel in the settings about toping up with an "add credits" button that brings up that dialog.

Likewise auto-top shoulld be a dialog and it should look like this:

<img width="467" height="521" alt="Image" src="https://github.com/user-attachments/assets/f5c4a1dd-dbfa-4537-b794-6e1f8219d571" />

Don't worry about the monthly limit. Likewise getting to auto-top should be a button in the settings that says modify.

Also clicking the pricing in the sidebar should bring up the add credits modal not the billing gate

PR #1356, opened 2026-08-27
 feat: improve low balance credit prompts

Jyatin:
## Summary

- Add a low-balance warning when credits decrease below the warning threshold.
- Show a clear `$0` balance warning when generation is disabled due to no credits.
- Provide direct actions from the warning toast to add credits or view other billing options.
- Make the credit balance pill and credit-related actions open the Add Credits dialog consistently.

## Validation

- `bunx tsc --noEmit` ✅
- Pre-commit checks passed:
  - dead-code ✅
  - format ✅
  - lint ✅
  - typecheck ✅

## Notes

The server-side billing/preflight checks remain authoritative; these changes improve the client-side credit UX and prompts.

### DIFF
--- a/src/hooks/use-low-balance-warning.ts
+++ b/src/hooks/use-low-balance-warning.ts
@@ -1,10 +1,6 @@
-/**
- * Low Balance Warning Hook
- * Fires toast notifications when balance decreases and crosses threshold
- */
-
 import { useEffect, useRef } from 'react';
 import { toast } from 'sonner';
+import { openAddCreditsDialog } from './use-add-credits-dialog';
 import { openBillingGate } from './use-billing-gate-dialog';
 import { useBillingBalance } from './use-billing-balance';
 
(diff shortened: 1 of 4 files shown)


src/hooks/use-low-balance-warning.ts
```diff
@@ -20,38 +16,38 @@ export function useLowBalanceWarning() {
...
+          onClick: () => openBillingGate(),
         },
         duration: 10_000,
       });
     } else if (isLowBalance) {
       hasWarnedRef.current = true;
-      toast.warning(`Balance is below $${lowBalanceThreshold}`, {
-        description: `Your balance is $${balance.toFixed(2)}.`,
+      toast.warning(`Balance is $${balance.toFixed(2)}`, {
+        description: 'Another short is about $13.',
```
tombeckenham: **[suggestion]** The new low-balance description hardcodes `'Another short is about $13.'`, and both toasts hardcode the action label `'Add $10'`. Those figures already exist as `TYPICAL_SHORT_COST_USD` (13) and `MIN_TOPUP_AMOUNT_USD` (10) in `src/lib/billing/constants.ts`. The `$13` line is the same “about ~$13, not a live estimate” copy the ready-email already interpolates from `TYPICAL_SHORT_COST_USD` (`src/lib/emails/notify-sequence-ready.ts`). Leaving literals here will drift the next time signup-grant / typical-short / min-top-up is retuned (the grant was already raised from $10 in #1140 because a default short outgrew it). The two numbers also sit awkwardly in one toast: the prompt says another short is ~$13 while the CTA offers $10, which is the minimum purchase, not enough on its own if the remaining balance is low. Clicking the action also does not add $10 — it only opens `AddCreditsDialog`, whose default `'10'` is itself a separate hardcoded string and is not reset on open (globally mounted, so a previously typed amount can stick).

**Suggestion:** Import `MIN_TOPUP_AMOUNT_USD` and `TYPICAL_SHORT_COST_USD` and interpolate both strings from those constants (same pattern as the ready-email). Prefer a label like `Add credits` unless `openAddCreditsDialog` actually seeds the dialog amount; if you keep “Add $N”, drive N from `MIN_TOPUP_AMOUNT_USD` and reset/pass that amount when opening.

src/hooks/use-low-balance-warning.ts
```diff
@@ -20,38 +16,38 @@ export function useLowBalanceWarning() {
...
         action: {
```
tombeckenham: **[suggestion]** The $0 toast correctly keeps a secondary “Other options” → `openBillingGate()` path (BYOK, founder grant, gift codes). The low-balance toast no longer does — it only offers “Add $10” → add-credits dialog. That undoes the #1099 rationale the PR deleted: both warnings used to open the gate so purchase was not the only exit. This is a worse fit for (a) fal BYOK / auto-top-up teams, who still get this hook (the sidebar pill already suppresses itself when `hasAutoTopUp || hasFalKey`, this hook does not) and (b) Stripe-disabled local/self-host, where `AddCreditsDialog` continue throws “Stripe is not configured” while the gate already hides the purchase card. The PR body also claims both add-credits and other billing options from “the warning toast”; only the $0 toast actually does that.

**Suggestion:** Mirror the $0 `cancel: { label: 'Other options', onClick: () => openBillingGate() }` on the low-balance toast, or skip/soften the toast when `stripeEnabled` is false or a safety net (fal key / auto-top-up) already covers generation — same idea as `CreditBalancePill`. If the $0 description stays “Generation is off until you add credits.”, that copy is also wrong for BYOK; the gate is the accurate next step.

tombeckenham:
Hi @Jyatin, thanks for raising this. There's a couple of good ideas on this, but I've already implemented some of this on other PRs. I'm going to add the 2 button option on the sonner though. I'm going to close this particular Pr, but feel free to take on any other issue. There's a lot!

PR #1394, opened 2026-08-31
fix(#1393): keep Generate live on an empty studio prompt

### LINKED ISSUE #1393: Keep Generate highlighted when prompt is empty; route empty clicks to login or a random prompt
## Why

30-day funnel: 1,667 visitors → 246 sign-in starts (15%) → 216 complete (88%). The login dialog converts fine — the leak is upstream. The studio composer's primary CTA is `disabled={!canSubmit}` (studio-composer.tsx), so every fresh visitor sees a dead Generate button.

## What

- Generate renders as an active primary button even when the prompt is empty (keep the placeholder text — it stays a placeholder, no pre-fill).
- Logged out + click Generate with empty prompt → open the login dialog (non-empty prompts already gate via `requireAuth`).
- Logged in + click Generate with empty prompt → alert/dialog: enter your own prompt, or **Try something random** — random uses the enhance flow to generate a script/prompt, then proceeds.
- Still disabled while uploads are in flight (`uploading > 0`) and when the selected mode isn't ready.

## Instrumentation

- `login_dialog_opened` with `source: 'auth_gate' | 'explicit'` in `auth-gate-provider.tsx` (requireAuth vs openLogin) — currently nothing records which button opened the dialog.
- Event for the empty-prompt Generate click and which option was chosen (own prompt vs random).

tombeckenham:
Closes #1393.

Both composers gated their primary CTA on non-empty text, so every fresh visitor met a dead Generate button.

**Studio composer** (`/images`, `/videos`)
- `canSubmit` is `modeReady && uploading === 0` — Generate is live on an empty prompt (the placeholder stays a placeholder, nothing is pre-filled).
- Empty + logged out → `requireAuth()` opens the login dialog.
- Empty + logged in → "What should we make?" with **I'll write it** / **Try something random**.
- Random goes through the **draft (enhance) flow**: the LLM writes the prompt from any attached references, or from nothing. It stops there — the prompt lands in the composer to be read and edited; generating it is still the user's click.

**Script composer** (`/`, `/sequences/new`)
- Generate no longer requires text — only a style pick and an analysis model. Empty + logged out → login dialog (intent remembered, so sign-in resumes into the dialog); empty + logged in → the same "What should we make?".
- Random runs the **enhance flow** in invent mode and stops: the script appears in the editor for the user to read, edit or undo before they spend anything on it.

**Instrumentation**
- `login_dialog_opened` with `source: 'auth_gate' | 'explicit'` in `auth-gate-provider.tsx`.
- `empty_prompt_generate_clicked` (`surface`, `authenticated`) and `empty_prompt_choice` (`surface`, `choice: 'own_prompt' | 'random'`).

**Tests** — `bun run test` (3397) and the e2e specs pass locally:
- anonymous empty Generate opens the login dialog, on `/images` and on `/` (`auth.spec.ts`)
- signed-in empty Generate opens the pick-a-prompt alert, on `/images` (`studio.spec.ts`) and on `/` via ⌘+Enter (`sequences.spec.ts`, which asserted the old disabled behaviour)

**Random invents; it doesn't borrow a sample**
- Script composer sends `invent: true` and the enhancer writes an original script — in the selected style, or in any genre when Match script is selected (no style block reaches the prompt). The instruction sits *outside* `<USER_SCRIPT>`, whose contents are narrative material by definition, so a "make something up" line in there would be written about rather than followed. The schema's 10-character floor lifts only for that flag; the public API's own schema is untouched.
- Studio drafts the same way — from the attached references, or from nothing.

**Six sample scripts per style, up from one**
- `generate-style-briefs.ts --variants N` asks for N genuinely distinct briefs in one call and writes `style-brief-variants.generated.ts`: 405 briefs across all 81 styles (canonical + 5 each). `max_tokens` scales with the ask — a truncated reply is unparseable JSON, which is what dropped the odd style out of a run.
- Samples stay **code, not data**: no `styles` column, no migration, resolved by style-name slug at runtime. A team's own custom style still falls back to its category brief.
- Shuffle and "Try this style" pick among canonical + variants, so touring a style twice no longer shows the same script. The `?style=` seed and the rendered sample videos stay pinned to the canonical brief (they have to be deterministic).


oudouusa:
The variant-grid de-flake commit `1af3fa82` is also needed by the VestaOS thin fork: its actions-less local certification reproduced the exact pre-hydration failure twice on upstream `070a7afb`, with the `Variant selection` grid missing after all Playwright retries (71 other tests passed each run). We are intentionally waiting for the upstream merge instead of cherry-picking the patch into the fork. All checks on this PR are currently green.

PR #1409, opened 2026-09-01
feat(#1408): stop-at generation stages with DAG continue

### LINKED ISSUE #1408: Workflow control
I want to give the user the choice of how far the workflow should run when clicking generate. I want the workflow to be able to stop and just about any point and still be recoverable. e.g. there should be some system in place that figues out from the DAG where we are up to in generation, then show the action button to generate that part next where the generate frames motion button appears. The dialog should mirror the progress bar that we show It should give the user a slider showing analysis, references, images... whatever the bar currently shows. Whatever the user selects is where the worlflow stops. I also want to remove the auto generate options from the options dialog, and remove model recomendation from styles. Styles should not be recomending models. Aspect ratio only

tombeckenham:
## Related Issue

Closes #1408

## Summary of Changes

Give Generate a **Run until** slider (Script → Casting → References → Images → Motion → Music) that mirrors the progress banner. The workflow stops at the selected stage, writes a checkpoint, and the scene-list footer offers the next DAG action so a partial run is recoverable.

- Auto-generate motion/music toggles are gone; stop-at is the per-run choice.
- Styles recommend **aspect ratio only** (no model badges or auto-apply).
- `analyze-script` returns after the requested stage; continue uses storyboard `resume` so existing shots are not wiped.
- Additive D1 columns: `generation_stop_at`, `pipeline_stage`, `generation_checkpoint`.

## Risk Assessment

- [x] Medium
- [ ] Low
- [ ] High

Partial pipelines are a new product state (scenes without stills, stills without motion). Credit preflight scales with stop-at so stopping at Script/Casting/References does not reserve a full render envelope.

## Additional Notes

Default stop-at remains **Music** (the short-film aha). API v1 `motion`/`music` flags still parse and map onto a stop-at stage.


tombeckenham:
Superseded by #1428 — GitHub refused to reopen this one after the branch was force-pushed past the commit it was closed at. Same branch, same work.

PR #1426, opened 2026-09-01
fix(#1418): document the real minimum Bun version (>= 1.3.9)

### LINKED ISSUE #1418: `bun dev` silently exits without starting the server on Bun < 1.3.9 (README says >= 1.3.0)
### Summary

Following the README Quick Start on a fresh clone, `bun dev` exits with code 0
without ever starting the dev server. Nothing listens on :3000. There is no
error message, so it reads as a broken repo rather than a toolchain problem.

The cause is the documented minimum Bun version. README and CONTRIBUTING both
state **Bun >= 1.3.0**, but `dev:app` depends on Bun's multi-script runner:

```json
"dev:app": "bun --sequential db:migrate:local db:seed:local dev:vite",
"dev:all": "... bun run --parallel dev:app stripe:dev dev:bunny",
```
`bun run --parallel / --sequential` were introduced in
[Bun v1.3.9](https://bun.com/blog/bun-v1.3.9). On anything older Bun does not
recognize the flag, so rather than running three scripts in order it runs only
the first and passes the remaining two as arguments to it.

**Steps to reproduce**
Install Bun 1.3.8 (the newest release that still satisfies the documented >= 1.3.0)
git clone the repo, bun install
bun dev

**Environment**
OS: macOS (Darwin 25.6.0), arm64
Bun: 1.3.8 (broken) → 1.4.0 after bun upgrade (works)
Node: v22.22.0
Browser: n/a — the server never starts

**Notes**
Nothing else about the setup was at fault — .env.local generation, migrations,
and seeding all worked once invoked directly.

CI pins bun-version: 1.3.14 in .github/workflows/test.yml, which is why this
has never been caught there. It only affects contributors installing Bun fresh
or on an older install — precisely the Quick Start path.

**Suggested fix**
Raise the documented minimum from 1.3.0 to 1.3.9 in the three places it appears:

the Bun badge in README.md
Prerequisites in README.md
Prerequisites in CONTRIBUTING.md

ai-shreyas:
## Related Issue

Closes #1418

## Summary of Changes

Raises the documented minimum Bun version from `1.3.0` to `1.3.9` in the three
places it is stated, plus the `engines` field:

- README badge
- README → Prerequisites
- CONTRIBUTING → Prerequisites
- `package.json` → `engines.bun`

`bun dev` does not work below 1.3.9. `dev:app` is built on Bun's multi-script
runner:

```json
"dev:app": "bun --sequential db:migrate:local db:seed:local dev:vite",
"dev:all": "... bun run --parallel dev:app stripe:dev dev:bunny",
```

`bun run --parallel / --sequential` were introduced in
[Bun v1.3.9](https://bun.com/blog/bun-v1.3.9). On older versions Bun does not
recognize the flag, so instead of running three scripts in order it runs the
first one and passes the rest as arguments to it.

## Risk Assessment
- [x] Low
- [ ] Medium
- [ ] High

## Additional Notes
Why CI never caught this. Every job in .github/workflows/test.yml pins
bun-version: 1.3.14, so CI has always run above the threshold — the broken
configuration never existed on a runner. The docs meanwhile advertised
>= 1.3.0, leaving 1.3.0–1.3.8 claimed-as-supported and tested by nobody. CI
also never invokes bun dev (Playwright starts the server via its own
webServer config), and the failure exits 0, so even a job that did run it
would go green.

Deliberately left out of this PR: I considered adding a Bun version check to
scripts/ensure-env.ts, which is the first thing bun dev runs — that would
turn the silent exit into an actionable error for people who don't re-read the
prerequisites. It's a behavior change on everyone's dev startup path, so I'd
rather have your call on whether you want it before writing it. Happy to open a
follow-up.


tombeckenham (approved):

ai-shreyas:
Thanks Tom for approving the changes.

The PR is approved, all checks are green, and the change is running on staging with no issues. Could you merge it when you get a chance? Thanks!

Also, What is your thought on this?

**scripts/ensure-env.ts, which is the first thing bun dev runs — that would
turn the silent exit into an actionable error for people who don't re-read the
prerequisites. It's a behavior change on everyone's dev startup path, so I'd
rather have your call on whether you want it before writing it. Happy to open a
follow-up.**

tombeckenham:
Thanks. Explain that bit further? I've never been sure about that script.

ai-shreyas:
I will check and see if that is helpful and required for this repository.

ai-shreyas:
> Thanks. Explain that bit further? I've never been sure about that script.

I had a read of the architecture and this script runs before actually runnign the code to ensure environemtn variables are loaded. 

 ```
From package.json
"dev": "bun scripts/ensure-env.ts && bun run dev:app",
  "dev:all": "bun scripts/ensure-env.ts && CLOUDFLARE_INCLUDE_PROCESS_ENV=true VIDEO_EXPORT_DEV_URL=http://localhost:8080 bun run --parallel dev:app stripe:dev dev:bunny",
```

I have made a PR #1470 for the fix.

PR #1470, opened 2026-09-03
feat(#1469): fail fast when Bun is older than the engines floor

### LINKED ISSUE #1469: Fail fast when Bun is older than the engines floor, instead of exiting silently
### Problem

`bun dev` fails silently on a Bun older than 1.3.9. It migrates the database,
exits 0, and never starts the server — no error, no non-zero exit:

`dev:app` uses `bun run --sequential`, added in
[Bun v1.3.9](https://bun.com/blog/bun-v1.3.9). Older versions don't recognise
the flag, so instead of running the three scripts in order they run the first
and pass the rest to it as arguments.

#1418 / #1426 raised the documented floor from 1.3.0 to 1.3.9, which fixes this
for anyone who reads the prerequisites and checks their version. It does not
help the case that actually bites: a contributor who *already* has an older Bun
installed, skims Quick Start, runs `bun dev`, and sees a clean exit with no app.
The repo looks broken rather than the toolchain looking old.

### Why documentation alone isn't enough

- The floor is stated in prose; nothing enforces it at runtime.
- `engines.bun` is advisory — no version manager reads it, and it installs
  nothing.

Only a check inside the process catches everyone, regardless of how their
runtime got onto the machine.

### Proposal

Add a preflight assertion to `scripts/ensure-env.ts`, which is already the
first thing both `bun dev` and `bun dev:all` execute:

```json
"dev":     "bun scripts/ensure-env.ts && bun run dev:app",
"dev:all": "bun scripts/ensure-env.ts && ... bun run --parallel ..."
```

I have made a fix in the ensure-env.ts file which will show this message when the bun version is older than required. This screenshot is just an example of what happens when there is requirement mismatch. 

Image

ai-shreyas:
## Related Issue

Closes #1469

## Summary of Changes

Adds a Bun version preflight to `scripts/ensure-env.ts`, the first thing both
`bun dev` and `bun dev:all` execute. On a Bun below the `engines.bun` floor it
prints the reason and the fix, then exits 1 — instead of the current silent
exit 0.

Follow-up to #1418 / #1426, which raised the documented floor. Documentation
helps someone who reads the prerequisites and checks their version; it doesn't
help someone who already has an older Bun installed and just runs `bun dev`.

Design notes:

- **The floor is read from `package.json` `engines.bun`**, never hardcoded — a
  second copy of the version would be the same drift bug in a new place.
- **No startup cost.** `process.versions.bun` is a property read, not a
  subprocess. Verified under the exact `bun scripts/foo.ts` invocation the
  package scripts use.
- **Fails open.** If the process wasn't launched by Bun, or `engines` can't be
  parsed, it returns without blocking. A preflight that halts development over
  an unreadable manifest is worse than the bug it prevents.
- **Covers `dev:all` too.** Rewriting `dev:app` with `&&` would fix `bun dev`
  on any Bun version, but `dev:all` uses `--parallel`, which `&&` can't
  express without a new dependency.

## Risk Assessment

- [x] Low
- [ ] Medium
- [ ] High

`ensure-env.ts` appears only in the `dev` and `dev:all` scripts — nothing in
build, deploy, or CI runs it. The blast radius is local development, and both
guard clauses fail open.

## Additional Notes

Verified locally:

- `oxfmt --check`, `oxlint --type-aware`, `tsgo --noEmit`, `bun dead-code` — all pass
- Guard **fires**: with `engines.bun` temporarily set above the installed
  version, exits 1 with the message above
- Guard **passes**: exits 0 on a supported Bun
- Full `bun dev` end to end: migrate → seed → `VITE ready` → HTTP 200 on :3000

## Screenshots.

If requirement mismatch, developer will see this on their terminal

Screenshot 2026-09-03 at 12 03 40 pm


tombeckenham (dismissed):

ai-shreyas:
All checks passed. @tombeckenham

tombeckenham (approved):

ai-shreyas:
Thanks. @tombeckenham

PR #1507, opened 2026-09-05
fix(#1506): show Omni video in PostHog generations with cost and via

### LINKED ISSUE #1506: Omni unsupported data type
## Context

Google Omni 1.1 Flash (`gemini_omni_flash` via native Google) may already be generating. The gap is **observability**, plus a pricing check.

Originally reported as an “unsupported data type” error. Production traces from 2026-09-04 show native Google successes (~$1.02 / 10s clip) and fal-via Omni failures (504s, safety). No “unsupported data type” events in the last 3 days.

## What to fix

1. **PostHog Generations / cost column**
   - Media spans use `gen_ai.operation.name=video_generation`. PostHog only classifies `chat` as `$ai_generation`, so Omni lands as an empty `$ai_trace` (no nested generation, `$ai_total_cost_usd` null).
   - Send a generation-classified span and `$ai_total_cost_usd` (our figure — PostHog cannot price `gemini_omni_flash` from OpenRouter tables).

2. **Failure via is hardcoded `fal`**
   - `MotionWorkflow.onFailure` always records `provider: 'fal'`, so native Google failures are invisible when filtering by Google.

3. **Studio video never records a span**
   - Sequence motion records after poll; studio video does not.

4. **Pricing check**
   - Video output $17.50/1M at 5,792 tokens/s of 720p is correct.
   - Input is $1.50/1M (text/image/video/audio) and is not billed today.

## Out of scope unless it reproduces

Request-shape / “unsupported data type” — treat as unconfirmed.

tombeckenham:
Fixes #1506

Native Google Omni 1.1 was generating (production traces from 2026-09-04 succeeded at ~\$1.02 / 10s). It did not show up usefully in PostHog AI observability.

**Observability**
- Media spans used `gen_ai.operation.name=video_generation`. PostHog only classifies `chat` as `\$ai_generation`, so Omni landed as empty `\$ai_trace` rows (no nested generation, `\$ai_total_cost_usd` null, invisible on the Generations tab). Spans now export as `chat` and set `\$ai_total_cost_usd` / `\$ai_output_cost_usd`. The duration histogram still uses `video_generation`.
- `MotionWorkflow.onFailure` hardcoded `provider: 'fal'`. Failures now record the resolved via (Google when a Google key is present).
- Studio video never recorded a span. Success and failure now do.

**Pricing**
- Video output \$17.50/1M at 5,792 tokens/s of 720p was already correct.
- Input is \$1.50/1M and was not billed; settled charges now include prompt tokens.

The original “unsupported data type” report did not reproduce in the last 3 days of production traces (failures were fal 504s and safety blocks).

### DIFF
--- a/src/lib/observability/ai-otel.ts
+++ b/src/lib/observability/ai-otel.ts
@@ -475,6 +475,15 @@ export type MediaActivity = 'video' | 'image' | 'audio';
  */
 export type MediaErrorType = 'content_filter' | 'provider_error';
 
+/**
+ * PostHog's OTLP mapper only classifies `gen_ai.operation.name=chat` as an
+ * `$ai_generation` event. Media semconv names (`video_generation` etc.) land
+ * as empty `$ai_trace` rows: no nested generation, `$ai_total_cost_usd`
+ * null, invisible on the Generations tab. The duration histogram below keeps
+ * the real activity so video latency is not mixed into chat.
+ */
+const POSTHOG_GENERATION_OPERATION = 'chat';
+
 export type MediaGenerationRecord = AIObservabilityMeta & {
   /** Model id as submitted to the provider. */
   model: string;
@@ -499,6 +508,10 @@ export type MediaGenerationRecord = AIObservabilityMeta & {
   costMicros?: Microdollars;
   /** Provider-reported billable units for the completed job. */
   unitsBilled?: number;
+  /** Provider-reported prompt tokens, when the via bills in tokens. */
+  inputTokens?: number;
+  /** Provider-reported completion tokens, when the via bills in tokens. */
+  outputTokens?: number;
   /**
    * True when a team's own provider key paid for this. `costMicros` is priced
    * the same either way, but only the `false` case is spend on us — without
@@ -569,6 +582,10 @@ function emitMediaGenerationSpan(record: MediaGenerationRecord): void {
     typeof record.outputUrl === 'string'
       ? [record.outputUrl]
       : record.outputUrl;
+  const costUsd =
+    record.costMicros !== undefined
+      ? microsToUsd(record.costMicros)
+      : undefined;
   const span = active.tracer.startSpan(
     record.observationName ?? `${operationName} ${record.model}`,
     {
@@ -576,10 +593,20 @@ function emitMediaGenerationSpan(record: MediaGenerationRecord): void {
       startTime: endTime - (durationMs ?? 0),
       attributes: {
         'gen_ai.system': record.provider,
-        'gen_ai.operation.name': operationName,
+        'gen_ai.operation.name': POSTHOG_GENERATION_OPERATION,
         'gen_ai.request.model': record.model,
-        ...(record.costMicros !== undefined && {
-          'gen_ai.usage.cost': microsToUsd(record.costMicros),
+        ...(costUsd !== undefined && {
+          'gen_ai.usage.cost': costUsd,
+          // PostHog's cost column reads these, not gen_ai.usage.cost, and
+          // cannot look up gemini_omni_flash in OpenRouter's price table.
+          $ai_total_cost_usd: costUsd,
+          $ai_output_cost_usd: costUsd,
+        }),
+        ...(record.inputTokens !== undefined && {
+          'gen_ai.usage.input_tokens': record.inputTokens,
+        }),
+        ...(record.outputTokens !== undefined && {
+          'gen_ai.usage.output_tokens': record.outputTokens,
         }),
         ...(record.unitsBilled !== undefined && {
           'tanstack.ai.usage.units_billed': record.unitsBilled,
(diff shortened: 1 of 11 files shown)


src/lib/workflows/motion-workflow.ts
```diff
@@ -1108,7 +1111,7 @@ export class MotionWorkflow extends OpenStoryWorkflowEntrypoint<MotionWorkflowIn
     // a step return this hook can't see.
     recordMediaGenerationSpan({
       model,
-      provider: 'fal',
+      provider: await resolveMotionVia(model, scopedDb.credentials),
```
tombeckenham: **[suggestion]** `onFailure` now `await`s `resolveMotionVia` (a D1 credential read) *before* `persistMotionFailure`. `recordMediaGenerationSpan` is carefully wrapped to never throw, but this await sits outside that guard. `emit-failure` runs `zeroOwnedReservation` then `onFailure` in one `step.do`; if via resolution throws, the whole cleanup step fails and the shot is never marked failed (healed only later by the reconciler). The live re-read can also disagree with the via actually used: `submitMotionJob` stamps `via: 'fal'` after a BytePlus portrait-filter fallback, but `onFailure` would still report `byteplus`.

**Suggestion:** Persist the failure first, then resolve via inside a try/catch defaulting to `'fal'` (or snapshot the stamped via somewhere `onFailure` can read). Do not let bookkeeping skip `persistMotionFailure`.

src/lib/workflows/studio-generation-workflow.ts
```diff
@@ -468,9 +491,31 @@ export class StudioGenerationWorkflow extends OpenStoryWorkflowEntrypoint<Studio
     error: string;
     scopedDb: WorkflowScopedDb;
   }): Promise<void> {
-    await scopedDb.generatedAssets.markFailed(event.payload.assetId, error);
+    const { assetId, userId, input } = event.payload;
+    if (input.activity === 'video') {
+      // Image failures are already recorded inside generateImageWithProvider.
+      recordMediaGenerationSpan({
+        model: input.videoModel,
+        provider: await resolveMotionVia(
```
tombeckenham: **[suggestion]** Same pattern as motion: `await resolveMotionVia(...)` runs before `generatedAssets.markFailed`. A credential-read failure now strands the reserved `generated_assets` row as not-failed, which this hook previously could not do. Success already records `provider: job.via`; failure should not be a more fragile path.

**Suggestion:** Call `markFailed` first, then emit the span with a swallowed via lookup (default `'fal'`).

src/lib/workflows/studio-generation-workflow.ts
```diff
@@ -434,6 +437,26 @@ export class StudioGenerationWorkflow extends OpenStoryWorkflowEntrypoint<Studio
       { url: videoUpload.url, contentType: videoUpload.contentType },
     ];
 
+    await step.do('record-video-observation', async () => {
+      recordMediaGenerationSpan({
```
tombeckenham: **[suggestion]** The new studio success span omits `durationMs`. `recordMediaGenerationSpan` then opens a zero-length span and skips the duration histogram (`if (durationMs === undefined) return`). Motion already stamps `submittedAt` on `MotionJobSubmission` and passes `Date.now() - job.submittedAt`; studio jobs have no equivalent, so studio Omni clips will show up in Generations with $0 latency.

**Suggestion:** Stamp `submittedAt` on `StudioVideoJobSubmission` the same way motion does, and pass `durationMs` here.

src/lib/observability/ai-otel.ts
```diff
@@ -569,17 +582,31 @@ function emitMediaGenerationSpan(record: MediaGenerationRecord): void {
...
+        'gen_ai.operation.name': POSTHOG_GENERATION_OPERATION,
         'gen_ai.request.model': record.model,
-        ...(record.costMicros !== undefined && {
-          'gen_ai.usage.cost': microsToUsd(record.costMicros),
+        ...(costUsd !== undefined && {
+          'gen_ai.usage.cost': costUsd,
+          // PostHog's cost column reads these, not gen_ai.usage.cost, and
+          // cannot look up gemini_omni_flash in OpenRouter's price table.
+          $ai_total_cost_usd: costUsd,
+          $ai_output_cost_usd: costUsd,
```
tombeckenham: **[suggestion]** `$ai_output_cost_usd` is set to the full blended charge, including the Omni input tokens this PR just started billing. PostHog's cost column can use `$ai_total_cost_usd` *or* input+output; with input unset and output equal to total the headline is fine, but any output-cost breakdown will count prompt/image tokens as video output.

**Suggestion:** For token-billed vias, set `$ai_input_cost_usd` / `$ai_output_cost_usd` to the actual split and keep `$ai_total_cost_usd` as the sum. Leave fal (no token split) as total-on-output.

Issue #1445, opened 2026-09-02
Enforce the client/server import boundary with lint rules, not just a test


tombeckenham:
## Summary

The client/server import boundary is real and well-specified, but it is enforced *only* by a test that runs after the fact. Nothing tells you at the point of the mistake. Add oxlint rules for the per-file shapes that cause the leaks, so the boundary is visible in the editor.

## What exists today

**`src/lib/client-server-boundary.test.ts` (#1253, #1257)** — the substantive guard, and a good one. It walks the transitive import graph from `src/components`, `src/hooks` and `src/functions`, and models how the Start compiler ships server fn files (`.handler(…)` / `.server(…)` arguments compiled out, then `deadCodeElimination` via `@tanstack/router-utils`). It fails if any client-reachable path reaches a `SERVER_ONLY` specifier — `stripe`, `#db-client`, `#storage`, `@tanstack/ai*`, `@opentelemetry/`, `@libsql/`, `drizzle-orm/libsql`, `better-auth`, `cloudflare:workers`. 8 tests, ~1.3s.

**`.oxlintrc.json` `no-restricted-imports`** — narrow by design: blocks `#db-client`, and `createScopedDb` from `@/lib/db/scoped`. That is the "database access only in server handlers" rule from CLAUDE.md, not an import boundary.

## The gap

The guard is invisible where the mistake is made:

- **No editor feedback.** You write the import, everything looks fine, and you find out later.
- **It does not run pre-commit.** `lefthook.yml` runs lint, typecheck, dead-code and format on staged files — not tests. So the boundary is a CI-time failure.
- **It is not discoverable.** Two people on this repo independently assumed a lint rule already enforced this. If the convention is only legible by reading a test file, it will keep being rediscovered the hard way.

The manual discipline it demands is real. Server fn files carry hand-written header comments explaining that only handler bodies may reference server-heavy modules; `src/lib/ai/fal-cost.ts` uses a type-only static import plus a dynamic import at the call site specifically to keep `fal-pricing-live`/drizzle out of the client graph. That reasoning should be mechanical, not remembered.

## What lint can and cannot do here

oxlint rules are **per-file** — they have no module graph. The transitive walk must stay a test; this issue does not propose replacing it. What lint *can* do is reject the local shapes that produce the leaks, instantly:

[...]

PR #1508, opened 2026-09-05
refactor(#1489): finish the client/server split — delete 19 of 20 !@/lib exceptions

### LINKED ISSUE #1489: Finish the client/server split: delete all 20 remaining !@/lib exceptions
Follow-up to #1445 / #1450. That PR established the boundary (`src/lib` server-only, `src/shared` client-safe), enforced it per-file with `no-restricted-imports` plus a transitive test, and moved the all-shared directories. It landed with **20 `!@/lib/<dir>` exceptions** for directories that still mix both halves.

This issue is deleting all 20 in one pass.

## Why it's worth doing as one hit

The exceptions aren't bookkeeping — they're the reason server code still ships to the browser. Measured against the current build via sourcemap attribution:

```
183 KB  lib/motion      ← 105 KB of it is generated/{schemas,zod}.gen.ts
167 KB  lib/ai          ← models.ts 29, input-hash 26, scene-analysis.schema 21
163 KB  lib/style
104 KB  lib/realtime
 40 KB  lib/prompts
 30 KB  lib/billing
 ... 
870 KB  TOTAL src/lib source still in the client bundle
```

(For scale: #1450 removed 360 KB of minified client assets by fixing one import form.)

Doing it in one pass is also cheaper than batching, because the dependency graph collapses — see the two waves below. Batching would mean repeatedly re-deriving the same graph.

## Current state: 20 exceptions, 360 value imports

Counts are what lint flags if you delete the line (type-only imports are already legal and don't justify an exception):

| dir | n | dir | n | dir | n | dir | n |
|---|---|---|---|---|---|---|---|
| ai | 58 | auth | 36 | style | 30 | billing | 29 |
| observability | 27 | shots | 20 | realtime | 19 | motion | 12 |
| studio | 11 | scenes | 8 | compliance | 7 | talent | 6 |
| models | 6 | sequences | 5 | sequence-elements | 5 | image | 5 |
| sheets | 2 | workflows | 1 | storage | 1 | prompts | 1 |

## Wave 1 — 48 modules move as-is (215 of 360 imports, ~60%)

Every direct value import in these is already clean. Mechanical `git mv` + path rewrite:

`shots/shot-view`(28) `observability/logger`(27) `auth/session-query`(13) `billing/money`(11) `models/catalog`(10) `scenes/scene-selection`(9) `ai/models.config`(8) `realtime/client`(8) `auth/route-guards`(7) `billing/constants`(7) `scenes/scene-segments`(7) `style/style-assets`(6) `shots/use-start-frame`(5) `shots/update-stale-depth`(5) `talent/sheet-progress-copy`(5) `auth/client`(4) `ai/scene-analysis.schema`(3) `observability/player-events`(3) `realtime/generation-stream.reducer`(3) `sequence-elements/limits`(3) `sequences/list-prefs`(3) `studio/outputs`(3) + 26 more at ≤2 each.

## Wave 2 — 51 modules, but it collapses to a few roots

Most are blocked *only* by a Wave-1 module, so they fall out for free once Wave 1 lands. The genuine work:

**`ai/models` is the keystone (34 imports).** Blocked by `@/lib/ai/grok-native` + `@/lib/motion/endpoint-map` (the latter pulls `MOTION_INPUT_SCHEMAS` → the 105 KB of codegen'd fal schemas). At least 8 other modules are blocked solely by `ai/models` — `ai/generation-mode`(9), `studio/schema`(7), `ai/resolution-support`, `image/build-image-request`, `motion/build-model-input`, `sheets/sheet-image-model`, `shots/set-image-offer`, `billing/cost-estimation`(7).

Client code only wants catalog *data* from it — `IMAGE_MODELS`, `getAudioModelDurationLimits`, `AudioModel`. So this is a **split**, not a move: catalog (ids, labels, capabilities, ratios, duration limits) → `src/shared`; endpoint/adapter/pricing wiring stays in `src/lib`. `ai` and `motion` are one problem, not two — worth ~250 KB and unblocks the largest cluster.

**Small genuine splits:** `ai/content-rejection` ← `ai/fal-error`; `style/style-templates` ← `storage/public-assets`; `compliance/attestations` ← `compliance/hash`; `storage/upload-target` ← `auth/system-admin` + `db/scoped`.

## Do NOT migrate these two

Documented in `.oxlintrc.json`, repeating here because the natural instinct is wrong:

- **`!@/lib/storage`** — exists for `routes/r2.$.ts`, whose import sits inside a `server: { handlers }` block the client build prunes. `serve-media` reaches `#storage`; it can never move to `src/shared`. The exception is working around the per-file rule not seeing through the strip.
- **`!@/lib/workflows`** — one `.test.ts` import only. No production client file needs it.

`!@/lib/auth` is also narrower than it looks: all three `auth/config` importers are server routes. The exception is really for `session-query` / `route-guards` / `client` / `navigation`.

## Structural follow-ups worth folding in

- **A `src/browser/**` tier.** `src/shared` currently conflates "runs anywhere" with "runs only in a browser" — `sequence-player/export.ts` uses `AudioContext`/`OfflineAudioContext`, `voice/speech-recognition` wraps `webkitSpeechRecognition`. Nothing stops a Workers-side `src/lib` module or workflow importing them; it typechecks, it lints, and it dies at runtime. Given #968 exists precisely because WebCodecs isn't on Workers, this is a live hazard, and the current split makes the conflation permanent.
- **`tsconfig.shared.json`.** Once the exception list is empty, a second tsconfig for `src/shared` with no `@/lib/*` path mapping turns every violation into a compiler error (TS2307) instead of a lint rule — the invariant moves into the type system. Only viable after this issue is done.

## Definition of done

- [ ] Zero `!@/lib/<dir>` exceptions in `.oxlintrc.json` except the two documented non-migrations
- [ ] `src/lib` source in the client bundle materially down from 870 KB (verify by sourcemap attribution on `bun run build`)
- [ ] `bun lint`, `bun tsgo --noEmit`, `bun run test`, `bun dead-code` green
- [ ] CLAUDE.md's boundary section updated to match the final exception list

## Reproducing the numbers

Per-directory violation counts: strip every `"!@/lib/..."` line from the boundary pattern group in `.oxlintrc.json`, run `bun lint`, and group the `no-restricted-imports` hits by directory. Bundle attribution: `bunx vite build --sourcemap`, then sum `sourcesContent` lengths per `src/lib/<dir>` across `dist/client/assets/*.js.map`.

tombeckenham:
Closes #1489. Follow-up to #1445 / #1450.

## What changed

**19 of the 20 `!@/lib/<dir>` exceptions are gone.** Only `!@/lib/storage` remains (`routes/r2.$.ts`, handler-only import, documented in `.oxlintrc.json`). `!@/lib/workflows` existed for one `.test.ts` import; tests now join the server-route lint override, so that line was dead and is deleted too.

**106 modules (+ co-located tests, 182 files) moved from `src/lib` to the mirrored `src/shared` path.** The move set is the transitive closure of every lint violation, computed from the import graph, minus anything that reaches a server-only package. Every alias, relative, and textual reference (scripts, docs, configs, `vi.doMock` strings) was rewritten by script.

**Splits where a moving module still reached server code:**
- `ai/fal-cost`: `falCostFromUnits` reads live D1 pricing, so it moves to `lib/ai/fal-cost-billing.ts` with a static import. This replaces the dynamic-import hack; the pre-flight estimator stays client-safe in `shared/ai/fal-cost.ts`.
- `USER_UPLOAD_MODEL` → `shared/shots/user-upload-model.ts` (the only thing `set-image-offer` needed from `upload-media`).
- `formatReportReference` → `shared/compliance/report-reference.ts` (the only thing the moderation route needed from `provenance`).
- `auth/server.ts` (a bare `createServerFn`) → `functions/session.ts`, so `session-query` / `route-guards` live in `src/shared` with no lib edge.
- `workflows/scene-matching.ts` is pure (type-only db imports) → `shared/scenes/scene-matching.ts`.

**Two deliberate non-moves, handled in lint config:**
- `routes/[.]well-known/$.ts` and `routes/oauth/login.ts` are server-only routes (their only `createFileRoute` option is `server`, pruned from the client tree). They join the `routes/api/**` override rather than being exempted by directory.
- `shared/billing/billing-observability.ts` keeps its `@/lib/posthog-server` import inside `createIsomorphicFn().server(…)` under an `oxlint-disable-next-line` with the reason. The compiler strips it and the transitive test models that strip.

**One real leak the transitive test caught once the exceptions came off:** `functions/ai.ts` had `type X = z.infer<typeof styleRecommendationResponseSchema>`. A `typeof` in a type alias is a value reference, so the schema module survived DCE on the client. The inferred type is now exported from `response-schemas` and type-imported.

Codegen scripts (`motion-openapi-ts.config.ts`, `generate-motion-endpoint-map.ts`, `pull-motion-schemas.ts`) now write to `src/shared/motion/generated`.

## Bundle attribution (`bunx vite build --sourcemap`, `sourcesContent` summed per dir)

| | main | this branch |
|---|---|---|
| `src/lib` source in client bundle | 876 KB | **0 KB** |
| `src/shared` source in client bundle | 184 KB | 994 KB |
| all client sources | 18,178 KB | 18,114 KB |
| client JS on disk | 7,210 KB | 7,200 KB |

Be clear about what this is: the boundary is now enforced with no directory-level holes, and ~64 KB of genuinely server-only source (the fal billing half, response schemas) left the client. But most of the 876 KB **was not server code**. The request inspector in `scene-script-prompts.tsx` calls the real `buildMotionRequest` / `buildImageRequest` / native builders client-side, and `shot-duration-field.tsx` reads `MOTION_JSON_SCHEMAS`, so the 183 KB of generated fal schemas ship because the client uses them. The issue's proposed "catalog vs wiring" split of `ai/models` would not have changed that. The lazy honest outcome is a move, not a split, for that cluster.

## Not done here (follow-ups worth filing)

- **Compute the request-inspector preview server-side.** That is the only way the generated motion schemas leave the client (~180 KB source). It is a feature re-architecture (the preview is a `useMemo` over a lot of client state), out of scope for a boundary issue.
- **`src/browser/**` tier** (from the issue): `src/shared` still conflates "anywhere" with "browser only" (`sequence-player/export.ts`, `voice/speech-recognition`, and now `shared/realtime/*` hooks, `shared/auth/client`).
- **`tsconfig.shared.json`** with no `@/lib/*` path mapping is now viable, since the exception list is effectively empty.

## Checks

`bun lint`, `bun typecheck`, `bun run test` (4004 tests), `bun dead-code` all green; pre-commit ran the same plus the boundary test.

https://claude.ai/code/session_01T4J8H6v93kBjzoidgfDmQH

## Db access audit (second and third commits)

Asked while reviewing: is raw db access widely bypassed? Resolving every value import in `src` (alias, relative, bare), rather than trusting the lint patterns:

- The raw handle is read in exactly the allowlisted files: the scoped-db factory, the Better Auth adapter, the three pricing crons, the live pricing reader, and the two seeds. No server fn or route touches drizzle directly. Only the two request middlewares and the workflow base mint a `ScopedDb`.
- **What was loose was the config.** One override turned `no-restricted-imports` off for 12 files where each needed one of its two restrictions; one entry was dead; the message named 2 allowed files when 9 were; and the rule matches the literal specifier, so `from './client-d1'` walked past it.
- Now: one override per capability, and `src/lib/db/db-access-allowlist.test.ts` pins both allowlists by resolving imports (verified to fail on a relative-import probe). The boundary test's config check allows deliberate narrowing but still fails on drift.
- **Table SQL moved out of two allowlisted files.** The sign-up hook duplicated the `teams` + `team_members` bootstrap in `ensureUserAndTeam`; both now call `createDefaultTeam`. The e2e pricing seed moved from the workflow base into `seed-model-pricing`, so the base no longer holds the raw handle.

**Structural finding, not fixed here (worth its own issue):** the "scoped" db is team-scoped at the root domains (sequences, talent, styles, locations, library, billing, api keys, generated assets, compliance) but 18 child domains (shots, frames, variants, prompt versions, scenes, events, exports, characters, sequence locations/elements, sheet variants) take only the raw handle and contain no `teamId`. `scopedDb.shots.getById(id)` returns any team's shot. Isolation rests on each handler verifying the parent first. I audited all 253 server fns: 121 touch a child domain, 12 do so without `sequenceAccess`/`shotAccess` middleware, and every one of those 12 checks the parent through a root domain before the child call. So no hole found, but the guarantee is per-handler convention. The structural fix is child factories that take `teamId` and join through `sequences.teamId` (or an ownership check inside `getById`), which is a large change and out of scope for a boundary issue.


tombeckenham:
Superseded by #1536, which re-does this from main after #1516 landed (same 10 pre-move commits; the domain layout redone incrementally with a domain-blind platform).

Issue #1108, opened 2026-08-06
Manual pipeline driveability: full CRUD + DAG invalidation


tombeckenham:
## Summary

Make every sequence-pipeline entity fully CRUD-able so filmmakers (and later API/MCP) can drive the whole pipeline manually, with correct **downstream-only** invalidation.

## Why

1. **Creative flexibility** — add/edit/reorder scenes, shots, cast, locations; edit prompts; upload/replace stills and videos
2. **DAG proof** — mutating X only stales dependents of X; simultaneous prompt+image replace keeps image fresh
3. **Recovery** — after workflow fail/cancel, finish the pipeline manually or via Update all
4. **API/MCP foundation** — server-fn surface first; public API later (out of scope here)

## Decisions (locked)

| Decision | Choice |
|----------|--------|
| Delete | Soft-delete + restore; toast Undo only; no product hard delete |
| Upload kind | `frame_variants.kind = 'upload'` |
| Script → sheets | No — script only stales prompts |
| Multi-frame last/key | Defer |
| Phase order | **Media first** → cast/location → structure → polish |

## Phases

1. **Media inject** — upload/replace still + video; atomic `replaceFrameContentFn` (prompt+image)
2. **Cast/location** — bible edit + soft-remove
3. **Structure** — create/reorder/soft-delete scenes & shots
4. **Polish** — title, unarchive, music prompt after complete

## Out of scope

- Public API / MCP
- Hard delete UI
- Show-deleted trash list
- Script merge/split (#1037)

## Plan doc

See `docs/plans/manual-pipeline-crud.md` on the branch for this issue.

tombeckenham:
Follow-up for the dual-hash verify fallbacks: #1371 (delete after 2026-09-28). Catalog hashes are not nulled; verify accepts current or legacy digests until then.

Issue #1079, opened 2026-07-28
Edit element names


tombeckenham:
When uploading elements for a new script, the analyser picks names for what's uploaded. It would be useful it if took notice of the file name used (if any) and allowed me to edit the name before using. Then I also want to be able to select elements in my script before clicking generate. Nothing's highlighting.

PR #1509, opened 2026-09-05
fix(#1471): draft element uploads land at a permanent key

### LINKED ISSUE #1471: Drop the elements temp-then-move: draft uploads should land at a permanent key
### Symptom

Element thumbnails render as broken images (alt text only) in the "Upload reference elements" popover. Seen repeatedly across users, usually the whole grid at once with tokens already resolved (`DOMINICAN_JERSEY`, `RUNWAY_LOGO`, …) — so the images uploaded and vision-analysed fine. They 404 only afterwards, because the R2 object behind the URL is gone.

### Ruled out

Verified against production before filing:

- **Not CORS.** `` loads aren't CORS-gated, and `AppImage` emits a plain ``/`srcset` — no `crossorigin`, no fetch in the path.
- **Not the browser / not drag-and-drop.** Nothing in the path is browser- or input-method-specific.
- **Not the transform chain.** `https://assets.openstory.so/cdn-cgi/image/…/https://assets.openstory.so/styles/action/thumbnail.webp` → `200`, `cf-resized: internal=ok`. `onerror=redirect` returns `307` to the original on a missing source, and Cloudflare *does* follow the `/r2/<key>` → `storage.openstory.so` 302 when fetching a source (tested through an on-zone redirect: `200`, `cf-resized: internal=ok`).

### Root cause: the temp-then-move design

Draft elements upload to `elements/<teamId>/temp/<uploadId>.<ext>` (`src/functions/sequence-elements.ts:75`) purely because there is no `sequenceId` yet — the permanent path is `elements/<teamId>/<sequenceId>/<id>.<ext>`. On creation, `promoteTempElements` **moves** each object: `copyFile` then `deleteFile` (`promote-temp-elements.ts:81` → `storage-cloudflare.ts:250`).

That reason doesn't survive multi-model creation: one draft upload has to serve N sequences, so *move* is the wrong verb regardless. It produces two failure modes.

**A. Stale draft URLs after promotion** — the mode consistent with what users report. `promoteTempElements` deletes the temp object while `draftElements` (React state, `script-view.tsx:395`) still holds `tempPublicUrl`. `clearDraft()` on success clears localStorage but not that state, so the create view keeps rendering dead URLs until the user navigates or reloads. Same for a second tab holding the same restored draft (#1079). The deletion here is *correct*; the UI is what's wrong.

**B. Sequences race for the same temp objects** — reachable, but not yet observed in the wild; this is a code-read finding, not a reproduction. `createSequences` creates one sequence per selected analysis model, in parallel:

```ts
// src/lib/sequences/create-sequences.ts:264
const created = await Promise.all(
  analysisModels.map(async (modelId) => {
    …
    await promoteTempElements({ …, uploads: elementUploads });
```

Each call moves the *same* temp keys. The first to finish deletes them; the others hit `Source file not found` (`copyFile` throws, `storage-cloudflare.ts:271`), aborting that sequence inside `releaseReservationOnThrow` — and the objects are gone for everyone. Multi-select is live: `singleSelectAnalysis` defaults to `false` (`generation-settings.tsx:135`) and no call site sets it.

### Proposed fix: delete the temp concept

Rather than swapping `moveFile` for `copyFile` plus a sweep:

1. Draft uploads land at a permanent, sequence-agnostic key — `elements/<teamId>/uploads/<uploadId>.<ext>`.
2. Every `sequence_elements` row simply points at it. Rows are per-sequence; the object is shared and immutable.
3. `promoteTempElements` stops moving anything — it only inserts rows and triggers vision.

Kills the race, the dangling-URL class, and the `moveFile` call, in a smaller diff than copy-plus-sweep. Cost is orphan objects from abandoned drafts — a bucket lifecycle rule, or nothing at all given these are small reference images.

Fixing (1)–(3) makes mode A harmless too, since the URL stays valid after creation. Clearing `draftElements` alongside `clearDraft()` is still worth doing on its own merits.

### Repro (mode B)

1. New sequence, upload 2+ reference elements, wait for tokens to resolve.
2. Select **two or more** analysis models.
3. Generate.
4. Expect: both sequences carry all elements. Actual: one gets them, the other errors on promotion, and the temp objects are gone.

### Follow-up

Count existing `sequence_elements` rows whose `imageUrl` has no object behind it — those are permanently broken and want a backfill or a UI fallback, separately from this fix.

AdemBenAbdallah:
Closes #1471.

## The bug

Draft element uploads landed at `elements/<teamId>/temp/<uploadId>.<ext>` and `promoteTempElements` **moved** them to `elements/<teamId>/<sequenceId>/…` — `copyFile` then `deleteFile`.

Move is the wrong verb: one draft upload has to serve N sequences, because creation fans out one sequence per selected analysis model (`singleSelectAnalysis` defaults to `false` and no call site sets it). Two failures fall out of that:

**A. Stale draft URLs after promotion** — the reported symptom. Promotion deletes the temp object while `draftElements` still holds `tempPublicUrl`, so the "Upload reference elements" grid renders broken images with the tokens already resolved (`DOMINICAN_JERSEY`, `RUNWAY_LOGO`, …) until the user navigates or reloads. Same for a second tab holding the same restored draft (#1079).

**B. Sequences race for the same objects** — `createSequences` promotes inside a `Promise.all` over the selected analysis models, each call moving the *same* keys. The first to finish deletes them; the others throw `Source file not found` inside `releaseReservationOnThrow`, and the images are gone for everyone.

## The fix — delete the temp concept for elements

Rather than swapping `moveFile` for `copyFile` plus a sweep:

- **Draft uploads presign to `elements/<teamId>/uploads/<uploadId>.<ext>`** — permanent and sequence-agnostic. Rows are per-sequence; the object is shared and immutable.
- **`promoteTempElements` moves nothing.** It validates the client-supplied path is inside the team's namespace, re-derives the public URL from that path (rather than trusting the payload), and inserts the row pointing straight at it. Paths under the old `temp/` prefix still promote, so drafts already sitting in a browser's localStorage keep working.
- **The public API's element ingestion lands under `uploads/` too** (`ingestImageToTempBucket` gained a `pathPrefix`), so a caller-hosted image isn't left permanently live under a name a future temp sweep would reclaim. Talent and locations still ingest to `temp/` — their create cores really do move the object.
- `isValidElementStoragePath` moves to `@/lib/sequence-elements/storage-path` so the server fns and promotion share one boundary check.

Net: one `moveFile` call and the whole dangling-URL class removed, in a smaller diff than copy-plus-sweep.

## Notes

- **Sharing an R2 object across sequences is safe here:** element product-delete has been soft since #1108, and nothing in the codebase deletes element bytes. `copySequenceElements` already *copies* (never moves) and is untouched.
- **The `tempPath` / `tempPublicUrl` wire names stay.** A draft is persisted to localStorage under those keys — renaming them would silently drop the elements out of every draft already saved in a browser. There's a comment on the schema saying so.
- **Cost:** orphan objects from abandoned drafts. Small reference images; a bucket lifecycle rule handles them, or nothing at all.
- **Not done here:** clearing `draftElements` alongside `clearDraft()`. With permanent keys the URL stays valid after creation, so mode A is fixed without it — and the draft-persist effect in `script-view.tsx` lists `draftElements` in its deps, so setting it to `[]` right after `clearDraft()` would re-save the draft it just cleared. Worth its own change.

## Tests

- New `src/lib/sequence-elements/promote-temp-elements.test.ts` (the module had none): the row points at the uploaded object, the URL is derived from the path and not the payload, **two sequences can claim the same upload** (the mode-B race), a legacy `temp/` draft still promotes, out-of-namespace paths are skipped, and the inline-vision / vision-workflow split is preserved.
- `storage-path.test.ts` carries over the existing traversal cases with the helper.
- `resolve.test.ts` updated for the `uploads/` prefix.
- `bun run test` — 4012 passed / 384 files. `bun typecheck`, `bun lint`, `bun dead-code` clean.
- Full-pipeline e2e (`bun test:e2e:full`) fails locally at the post-redirect `Show the script` assertion — **it fails identically on unmodified `main`** in this environment, so it isn't attributable to this change. CI is the signal there.


tombeckenham (dismissed):
Made a couple of changes. We no longer need promote temp elements after this

tombeckenham:
Thank you so much for this @AdemBenAbdallah !

tombeckenham (approved):

Issue #1564, opened 2026-09-10
e2e: the full-sequence element fixture is never used by the script, so element URLs are never dereferenced


tombeckenham:
## The problem

`e2e/tests/full-sequence.spec.ts:285` uploads `e2e/fixtures/broadcast-mic.jpg` as a sequence element. The script it uploads it into (`:214`) is a coral lipstick ad. Nothing in that script calls for a studio microphone.

Elements are not bound by literal token matching — `scene-split-workflow.ts:360-373` hands the LLM a `- TOKEN: description` list and lets it choose which scenes use them. Given a lipstick ad and a condenser mic, it declines. The recorded fixtures confirm it:

- `recorded/openrouter/element-vision/element-vision.json` — vision runs on the mic and assigns the token `BROADCAST_MIC`.
- `BROADCAST_MIC` then appears **exactly once** across every file in `e2e/fixtures/recorded/`: inside the `<ELEMENTS>` block of the **input prompt** to `script-bibles` ("The following user-uploaded elements are available…").
- **Zero** occurrences in any `visual-prompts` or `motion-prompts` output.

That is the offer, not the use. The element reaches: presign → R2 PUT → vision → `sequence_elements` row → one line in a prompt. **Nothing downstream ever dereferences its `imageUrl`.**

## Why it matters

A dangling element URL is undetectable in this spec *in principle*, not merely unasserted. The spec's browser-error capture (`:160-175`) watches `pageerror`, console error/warning, `requestfailed` and 5xx — a 404 on an `<img>` is none of those, and no `<img>` is ever rendered for it anyway.

This is the class of bug #1471 was: promotion deleted the object while rows still pointed at it. That bug passed this spec before the fix, and a regression would pass it after.

Related: `selectRecordedPipelineModels` (`e2e/fixtures/test-utils.ts:179`) pins only Image and Motion models, and `analysisModels` falls back to `[DEFAULT_ANALYSIS_MODEL]` (`use-sequences.ts:244`). So N=1 analysis model, and the multi-model fan-out — the race that motivated #1471, where N sequences claim the same upload — is never exercised either.

## Proposed change

**Swap the fixture for something the script actually calls for.** The script names the lipstick three times, including a shot dedicated to it:

> *unboxes a coral lipstick and turns it slowly to camera* … **CLOSE ON THE TUBE** — *the coral bullet twists up and catches the light* … *The lipstick lands beside the brand mark on the concrete ledge*

[...]

PR #1565, opened 2026-09-10
fix(#1471): attach element uploads at a permanent key, one attach path

### LINKED ISSUE #1471: Drop the elements temp-then-move: draft uploads should land at a permanent key

tombeckenham:
Closes #1471. Supersedes #1509 — same diagnosis and same core fix, re-landed on the post-#1489 tree (that PR predates the restructure and conflicts on every file it touches). Original work by @AdemBenAbdallah.

## The bug

Draft element uploads landed at `elements/<teamId>/temp/<uploadId>.<ext>` and `promoteTempElements` **moved** them to `elements/<teamId>/<sequenceId>/…` — `copyFile` then `deleteFile`.

Move is the wrong verb: one draft upload has to serve N sequences, because creation fans out one sequence per selected analysis model (`create-sequences.ts`, `Promise.all(analysisModels.map(...))`). Two failures fall out:

- **Stale draft URLs after promotion** — the reported symptom. Promotion deletes the object while the create view still renders `tempPublicUrl`, so the "Upload reference elements" grid shows broken images with the tokens already resolved. Same for a second tab holding the same restored draft.
- **Sequences race for the same objects** — each of the N promotions moves the *same* keys. The first to finish deletes them; the rest throw `Source file not found` (`storage-cloudflare.ts:271`) inside `releaseReservationOnThrow`, and the images are gone for everyone.

## The fix

Draft uploads presign to `elements/<teamId>/uploads/<uploadId>.<ext>` — permanent and sequence-agnostic. Rows are per-sequence; the object is shared and nothing moves or deletes it.

**Sharing one R2 object across sequences is safe here, verified exhaustively:** every R2 mutation routes through `storage-cloudflare.ts`, and no `deleteFile`/`deleteFiles` caller ever passes `STORAGE_BUCKETS.ELEMENTS` (the four callers hit VIDEOS/THUMBNAILS, TALENT ×2, LOCATIONS). Element delete has been soft since #1108; the hard `sequenceElements.delete` and `sequences.delete` both have zero callers; no team-deletion flow; no cron touches storage; `copySequenceElements` uses `copyFile`, never `moveFile`.

## Removing the move merged two functions that had converged

`promoteTempElements` and `finalizeElementUploadFn` were doing the same job — down to a byte-identical private `triggerElementVision` in each file. They now share one `attachElementUpload` core, which forced three disagreements to resolve:

- **The stored URL is derived from the validated path**, never taken off the payload. `finalizeElementUploadFn` and `replaceSequenceElementFn` wrote the client's `publicUrl` verbatim, and `mediaUrlSchema` accepts any absolute http(s) URL — so `imageUrl` could point at a caller-chosen host while `imagePath` looked legitimate. `publicUrl` is gone from both wire payloads.
- **An out-of-namespace path throws** instead of being skipped with a `logger.warn`, which silently created a sequence with fewer elements than the user picked. `finalize` and `replace` already threw; promotion was the outlier. `releaseReservationOnThrow` already wraps the call, so the credit hold releases cleanly.
- **The R2 object must exist.** The deleted `moveFile` was the only thing proving it did; without a check a row can point at a permanent 404 that nothing surfaces until image generation fails hours later. `fileExists` was already in the storage API with zero callers.

## Schema dedupe, and a live API bug it fixed

Three copies of the draft-upload shape had drifted apart. One `draftElementUploadSchema` now, shared by the create-flow wire, the public API and attach.

The public API documents `elements[].token` as optional and derives one from the filename when absent — but `createSequenceSchema` **required** it, so every token-less API create 400'd at `createSequenceSchema.parse`. Fixed by construction, with a test pinning it.

## `ingestImageToTempBucket` → `ingestImageToBucket`

Now takes a required `IngestPrefix`: `'temp'` (talent/locations, which do move their object to a permanent key) or `'uploads'` (elements, which never move). Required rather than defaulted — the two have opposite durability contracts, and the reclaimable one would have been what a caller got by forgetting an argument.

## Notes

- **The `temp*` wire names stay.** A draft is persisted to localStorage under them and `readSequenceDraft` `safeParse`s the whole draft, so renaming one field silently discards the *entire* saved draft (script and style included), not just its elements. The warning now lives on `draftElementSchema` — the schema that actually persists — and points at the `:v1` key as the escape hatch.
- **Nothing sweeps `temp/` and nothing ever has** — three crons, none touching storage; no lifecycle rule anywhere. So the orphan cost from abandoned drafts is unchanged by this PR, not introduced by it. An R2 lifecycle rule is **not** the fix either: every attached element points into `uploads/` forever, so an age-based rule there would delete live objects. Reclaiming has to be row-driven (joining `sequence_elements.imagePath`), noted where someone would go looking.
- **Behaviour change worth a look:** a legacy draft in localStorage pointing at a *different* team's namespace now hard-fails the create instead of silently dropping that one element.

## Tests

`bun run test` — 4084 passed / 389 files. `bun typecheck`, `bun lint` (0 errors), `bun dead-code`, `bun format:check` all clean.

New `attach-element-upload.test.ts` covers: the row points at the uploaded object, the URL is derived not trusted, **two sequences claim the same upload** (the race), a legacy `temp/` draft still attaches, out-of-namespace throws, a missing object throws, the inline-vision / workflow split, and the vision-trigger failure marking the row `failed`.

`storage-path.test.ts` carries the traversal cases over and adds a **presign→validator round-trip guard**: the key `presignDraftElementUploadFn` mints via `buildR2Key` must be what `isValidElementStoragePath` accepts. If either side's shape drifts, every draft element would otherwise be silently rejected at create with nothing else to catch it.

**Known e2e gap, filed as #1564:** `full-sequence.spec.ts` uploads `broadcast-mic.jpg` into a lipstick-ad script, so the LLM never binds the element to a shot — `BROADCAST_MIC` appears exactly once across all recordings, in an *input* prompt, and never in a visual/motion prompt output. Nothing dereferences the element's URL, so this bug class is undetectable there in principle. Fixing it needs a re-record, so it's tracked separately.


tombeckenham:
Superseded by #1509, which now carries this work on the original author's branch.

Issue #1552, opened 2026-09-09
ElevenLabs provider: adapter, platform key, price card, e2e hook


tombeckenham:
## Why
Character voices are designed and synthesised on ElevenLabs. Nothing in the repo talks to it yet.

## Scope
- Add `@tanstack/ai-elevenlabs` (ships `elevenlabsSpeech` for TTS and bundles `@elevenlabs/elevenlabs-js`, which has the Voice Design / create-voice endpoints the adapter does not wrap).
- `ELEVENLABS_API_KEY` platform secret + `.env.example` + `ensure-env`. **Platform key only, no BYOK**: designed voices live in the account that created them, so a team key would not see voices we design (same shape as `ARK_API_KEY`).
- `ELEVENLABS_BASE_URL` hook so e2e can point at aimock; record fixtures for TTS and Voice Design.
- Static price card like `src/billing/byteplus-pricing.ts`: TTS per character, Voice Design per call, merged into the effective pricing map so estimates never gate on $0 and billing never records $0. Dated, advertised rates, `recordFalUsage: false`, unaudited like xAI/Google spend.
- A `resolveKey('elevenlabs')`-style credential hatch usable from a workflow step.

## Out of scope
Voice cloning from an uploaded sample. Realtime/agents.

PR #1584, opened 2026-09-11
feat(#1553): opt-in character voices via ElevenLabs Voice Design, copied to/from talent

### LINKED ISSUE #1553: Voice on characters: opt-in Voice Design in the bible workflow, copied to/from talent
## Why
A character needs one voice that survives every scene and every sequence. Characters are created dynamically from the script, so the voice is created there too, behind an opt-in.

## Scope
**Schema** (`bun db:generate`, no hand-written SQL):
- `characters.voiceId`, `characters.voiceDescription` (nullable: a character without a voice is a legitimate state, not an unknown).
- `characters.useVoice` nullable boolean = inherit the sequence setting (same shape as `shots.useStartFrame`).
- `sequences.generateVoices` boolean, default false, snapshotted at Generate like `generateStartFrames`.
- Same two voice columns on `talent`. Casting a talent copies its voice onto the character; saving a character to the library copies it to talent (the sheet flow with one more column).

**Workflow**: when `generateVoices` is on, the character bible workflow adds a step per *speaking* character: the LLM drafts a voice description from the bible, ElevenLabs Voice Design returns previews (stored in R2), the top preview is saved as a voice and its id stored. Previews cost no voice slot; only the saved voice does. Deleting the character/sequence/talent deletes the voice (ElevenLabs caps saved voices per account by plan, so never design per shot, never leak slots).

**UI**: "Voices" toggle in the Generate dialog (off by default). A Voice section on the character card: description, the previews to audition, regenerate, per-character off switch.

## Out of scope
Voice cloning from an upload. Assigning an existing library voice id by hand (cheap follow-up).

Depends on #1552.

tombeckenham:
Closes #1553. Depends on #1552 and #1560 (both merged). Supersedes #1583, which GitHub auto-closed when its base branch was deleted.

## What

- **Schema** (`ADD COLUMN` only): `characters.voiceId / voiceDescription / voicePreviews / useVoice` (nullable; `useVoice` NULL = inherit), `sequences.generateVoices` (default false), `talent.voiceId / voiceDescription`.
- **Generate dialog**: a "Voices" switch next to "Use start frames", off by default, snapshotted onto the sequence and the storyboard payload like `generateStartFrames`.
- **Workflow**: `CharacterBibleWorkflow` spawns a `CharacterVoiceWorkflow` child per *speaking* character (`speakingCharacterIds()`: a bible name sharing a token with a dialogue speaker cue) that resolves `usesVoice()` true and has no voice yet. The child LLM-drafts `voiceDescription` when empty (`phase/voice-design-chat`), runs Voice Design, parks the previews in R2, saves the first as a voice and stamps its id. Billed at `VOICE_DESIGN_COST` ($0.30/call) with provenance; the References credit gate counts speaking characters.
- **Slots never leak**: previews cost no slot, only the saved voice does. One voice id is shared by copy (talent → character at cast, character → talent at save-to-library), so it is freed only via `releaseVoiceIfUnreferenced` (`countVoiceReferences` over both tables) on character soft-delete, sequence archive, talent delete and regenerate. The character upsert keeps a voice the row already holds.
- **Character card**: Voice section with the per-character switch, the auditions (`<audio>`), Generate / Regenerate + cost; the description is a bible field.

## Notes

- Sequences are archived, not deleted, so archive is where the cast's voices are released (unarchive shows "No voice yet" and Regenerate works).
- A voice-child failure fails the run like a sheet failure does — voices were asked for.
- Not done: picking a non-first preview as the saved voice (the previews are stored with their `generatedVoiceId`, so it is a small follow-up), and Voice Design in aimock (e2e never turns voices on).

https://claude.ai/code/session_01P6TywSwr6rAAbM9vLJtkMU


tombeckenham:
Closing briefly so the preview D1 is torn down: the voices migration was regenerated on top of #1585 under a new name after the rebase, and the old preview database has the old name recorded. Reopening once cleanup finishes.

PR #1599, opened 2026-09-13
feat(#1554): TTS per dialogue line in the motion run, folded into the motion hash

### LINKED ISSUE #1554: Dialogue audio: TTS per line in the motion run, folded into the motion hash
## Why
Once a character has a voice, every dialogue line is synthesised with it and handed to the video model as a reference audio, so the voice is identical across shots and sessions.

## Scope
- A step in `MotionWorkflow` (before submit) that, for each dialogue line whose speaker has a voice, runs TTS (`eleven_v3`, tone string mapped to v3 audio tags) and stores the clip in R2 alongside the shot. One file per line, no concatenation (Workers has no ffmpeg; Seedance 2.5 takes up to 10 files / 30.2 s combined so per-line is fine).
- Voice ids + lines + TTS model fold into the motion-prompt content hash **only when a voice is present** (same rule as `usesStartFrame`), so editing a line or changing a voice re-stales the shot and no stored digest moves.
- Shot duration is raised to cover the audio (2.5 clips are 4–30 s).
- Every voiced line is snapshotted onto the workflow payload at trigger time; no mid-run D1 reads. Credentials via the named hatch.
- Preflight estimate reserves the TTS cost (static card).
- Audio clip ids are stamped on `VideoManifestEntry` and the motion `shot_prompt_versions` row (provenance, never inferred).
- Studio parity: the Optimised prompt JSON carries the audio refs so paste-into-Videos rebuilds them.

## Out of scope
Auditioning/regenerating a single line from the scene panel (follow-up).

Depends on #1552 and #1553.

tombeckenham:
## Why

Once a character has a designed ElevenLabs voice, every dialogue line is synthesised with it and handed to the video model as a reference audio, so the voice is identical across shots and sessions.

Fixes #1554.

## What

- **MotionWorkflow** runs `synthesize-dialogue-audio` before submit: one `eleven_v3` clip per voiced line (tone mapped to v3 audio tags), parked in R2, attached as `kind: 'audio'` refs. User-bound `voiceToken` elements are left alone. Models with no audio reference slot (Grok, Omni Flash, Kling) skip TTS.
- Voice ids + lines + TTS model fold into the **motion-prompt hash only when a voice is present** (same shape-stable trick as `usesStartFrame` / `referenceOnly`). Editing a line or changing a voice re-stales the shot; voiceless shots keep their stored digest.
- Shot duration is **raised** to cover the audio bed (Seedance 2.5 is 4–30 s). Combined audio under the min (1.8 s on 2.5) overflows to prose instead of 400ing the request.
- Voiced lines are **snapshotted at trigger** (single-shot, batch, storyboard, smart-retry, Update all). Credentials via `resolveKey('elevenlabs')`. No mid-run character reads.
- Preflight reserves TTS cost from the static card (real character count at trigger; ~200 chars/shot stand-in on storyboard before split).
- Provenance: `VideoManifestEntry.audioClipIds` and `shot_prompt_versions.audioClips`. The optimised-prompt JSON carries those audio refs for paste-into-Videos.

Out of scope (as specified): auditioning/regenerating a single line from the scene panel.

## Test plan

- [x] Unit tests for matching, tokens, tone tags, hash fold, duration raise, TTS estimate
- [x] Motion workflow tests still pass (TTS skipped when `voicedLines` is empty)
- [x] `bun typecheck` / lint / knip / migration-safety
- [ ] Generate a Seedance 2.5 shot with Voices on and confirm lip-sync uses the character voice across two shots
- [ ] Confirm a voiceless sequence’s stored motion-prompt hashes still match (no digest move)
- [ ] Confirm Grok / Omni Flash shots with voices do not call ElevenLabs TTS


tombeckenham:
Closing so preview D1 (`openstory-pr-1599`) is torn down. The audio_clips migrations were regenerated after rebase (old snapshots forked past film-length). Reopening once that D1 is gone so migrate runs on a fresh database.

tombeckenham:
Cannot reopen after the force-push (GitHub: branch was force-pushed or recreated). Continued in #1607. Preview D1 `openstory-pr-1599` was deleted with wrangler.

Issue #1519, opened 2026-09-07
Enable direct providers


tombeckenham:
Enable direct providers - google, spacexai, and byteplus on preview branches, and update pricing table

PR #1620, opened 2026-09-14
fix(#1531): per-run ACR leases and persistent slot reservations

### LINKED ISSUE #1531: BytePlus ACR pool: reservation races and shared-lease / terminal cleanup bugs (downstream fixes and tests)
## Summary

While integrating the BytePlus ACR asset pool introduced in #1361 / #1515 into a downstream fork, we identified issues with slot reservation concurrency, lease ownership, and cleanup on workflow termination. We have implemented fixes and regression tests in the fork and are sharing the findings here.

Reviewed revision: `main@7fba4ef72fb44e1d952acfb6193d985b4b299daf` (introduced in `f7a78e895913e51a6dc1c91fe06b99e9592a226b`). This report is based on code inspection and local ledger/workflow tests. **It is not a report of an observed production incident against the live BytePlus API.** External APIs were stubbed; actual account capacity and live API behavior have not been verified.

### 1. One job releases another job's lease on a shared asset

Relevant code: [claimSlot / releaseLeases](https://github.com/openstory-so/openstory/blob/7fba4ef72fb44e1d952acfb6193d985b4b299daf/src/lib/db/scoped/byteplus-assets.ts)

1. Register asset X, then have jobs A and B call `claimSlot` for the same identity.
2. Complete A while B is still polling its generation job, and call `releaseLeases([X])`.
3. With the pool at capacity, request a different asset Y.

**Actual:** There is only one lease expiry per identity. A's release sets it to zero, making X eligible for eviction while B still uses it. If X is selected, the pool can proceed to `DeleteAsset`.

**Expected:** X must remain protected while B has a valid lease, even after A completes. This also occurs when individual child jobs release their own assets; it does not require a parent to release its children's leases.

### 2. Reservations are not persisted, so multiple requests can claim the last free slot

Relevant code: `claimSlot` / `recordSlot` in the same ledger, and [ingestPooledAsset](https://github.com/openstory-so/openstory/blob/7fba4ef72fb44e1d952acfb6193d985b4b299daf/src/lib/ai/byteplus-asset-pool.ts)

1. Start with an empty ledger and `capacity: 1`.
2. Call `claimSlot` for X, but do not perform external ingestion or `recordSlot` yet.
3. Call `claimSlot` for Y.

**Actual:** Both calls return `reserved`. The free-slot branch checks the row count without inserting a reservation, so this ordering alone reproduces the problem. Concurrent ingestion can then let both requests proceed to `CreateAsset`. Concurrent misses for the same identity can also create duplicate external assets, with the subsequent upsert retaining only the last asset ID in the ledger.

**Expected:** Pending reservations must count toward capacity, and only one request may claim the last slot. An identity already being ingested must not trigger another external creation. The interval between eviction and recording the replacement also needs a persistent reservation.

### 3. Studio does not release leases when its workflow terminates

Relevant code: [StudioGenerationWorkflow](https://github.com/openstory-so/openstory/blob/7fba4ef72fb44e1d952acfb6193d985b4b299daf/src/lib/workflows/studio-generation-workflow.ts)

1. Acquire an ACR asset during Studio video generation.
2. Complete the workflow successfully, or fail it through `onFailure`.
3. Inspect the ledger lease.

**Actual:** Submission receives `assetLedger`, but neither the success path nor the failure path has matching cleanup. The lease continues to pin the slot until expiry.

**Expected:** A terminating workflow should release the leases it acquired.

### 4. Falling back from BytePlus to fal skips cleanup on success

Relevant code: [MotionWorkflow success path](https://github.com/openstory-so/openstory/blob/7fba4ef72fb44e1d952acfb6193d985b4b299daf/src/lib/workflows/motion-workflow.ts#L917-L927)

1. Acquire an ACR asset on the BytePlus path.
2. Trigger a fallback-eligible error in the subsequent generation request, switching to a fal job.
3. Complete the fal job successfully.

**Actual:** Success-path cleanup is guarded by `job.via === 'byteplus'`, so the previously acquired lease remains.

**Expected:** Cleanup should release leases owned by the workflow, independently of the final provider.

## Fix implemented in the downstream fork

The implementation follows the fork's own structure, so we would like to discuss the upstream approach rather than assume it should be ported unchanged. The relevant design is summarized here without requiring access to the fork:

- Separate slots, which hold assets and pending reservations, from leases keyed by `(identity, owner)`. Use a workflow instance ID prefixed by workflow type (`motion:` / `studio:`) as the owner.
- Use one D1 `batch` to perform capacity-checked persistent reservation, replacement of an eviction candidate with a reservation, and lease acquisition for resident assets. Count pending reservations toward capacity.
- Finalize an asset only when its reservation token matches; finalize the asset and add its lease in the same batch.
- Call `releaseOwner(owner)` on Motion / Studio success and failure, preserving other jobs' leases. Renew leases by owner while polling.
- Migrate and preserve existing ledger assets and unexpired leases.

There is also a separate failure boundary between the external API and the database. Upstream currently continues toward creation after `DeleteAsset` fails. In the fork, a deletion failure retains the previous asset ID and the reservation and prevents proceeding to creation. An indeterminate `CreateAsset` result also retains the reservation, preventing duplicate submission on retry.

This conservative handling has a limitation: **reservations with indeterminate external outcomes are not automatically reclaimed solely because time has elapsed; they require reconciliation with external state.** Fully automatic recovery is not implemented. An appropriate reclamation/reconciliation policy for upstream operations still needs discussion.

## Downstream validation

We added regression tests against the corrected implementation for:

- 12 concurrent requests for the final free slot, and 8 concurrent misses for the same identity plus retries.
- Releasing one owner of a shared asset while preserving the other owner's protection.
- Concurrent resident-asset acquisition and eviction.
- Preventing duplicate submission during ingestion, and retaining reservations after deletion failure or loss of a creation response.
- Legacy ledger migration and whole-batch rollback when lease insertion fails during finalization.
- 10 concurrent reservations, finalization, and owner-specific release against actual local D1 in Workerd / Miniflare.
- Owner-specific cleanup on Motion's fal success path and Studio's success/failure paths.

These are **regression tests of the downstream fixes**, not a claim that the entire test set was run against unmodified upstream code. The full unit suite and local E2E tests also passed, but they do not establish live BytePlus API behavior on either success or failure paths.

Environment: Linux / WSL2, Bun 1.3.14, Vitest, and local D1 through Miniflare / Workerd. These are server-side issues; the browser is not relevant to reproduction.

## Proposed upstream direction

Would per-job leases and persistent, atomic slot reservations be an appropriate basis for the fix, together with consistent Studio / fallback cleanup? If separate issues would be easier to track, these findings can be split by concern.

tombeckenham:
Closes #1531.

All four bugs from the report were still there after the #1519 refactor. The gap between claim and create is now wider, because a governor `step.sleep` sits between them.

## What changed

**1. One run could unpin another run's still.** Leases moved to a new `byteplus_asset_leases` table, one row per (still, run). The owner is `motion:<instanceId>` or `studio:<instanceId>`. `releaseLeases(identities)` is replaced by `releaseOwner(owner)`, which leaves other runs' leases alone. Every claim and finalize renews **all** of the run's leases, so a still leased early cannot expire while a later still waits the governor.

**2. Reservations weren't saved.** A reservation is now a `byteplus_assets` row with `asset_id` NULL, plus `reserved_by` and `reserved_until`.
- It counts against capacity. The free-slot path is one insert that checks capacity and does nothing on a duplicate still, so exactly one claimant gets the last slot.
- Eviction is one conditional UPDATE. It gives the victim's row to the new still, but only if no run holds a live lease on it and no row already has that still.
- A run that finds another run creating the same still gets `pending`. The claim step then retries (40 × 30s, ~20 min — sized for the governor queue plus create) and ends up with a hit, so the asset is created once.
- A full pool with every slot leased fails immediately (`NonRetryableError`). It does not wait out the 45-minute TTL; admission already waited, and 20 minutes cannot outlast a lease.
- A reservation that outlives the TTL can be taken over. Finalize records the Ark asset whoever holds the row (or inserts if the row is gone); it refuses only when the row already holds a *different* assetId.
- A replayed claim step finds its own reservation. A replayed create step re-finalizes the same asset without error.

**3. Studio never released its leases.** It now releases on success (step `release-byteplus-asset-leases`) and in `onFailure`.

**4. Motion's success release was Ark-only.** It now always runs, whichever via the clip finally rendered on. A success-path release that never lands is logged, not a failed clip (the TTL frees the stills). The `onFailure` release is last and uncaught, so `emit-failure` retries it.

**Eviction failures.** `DeleteAsset` is now its own `-evict` step. A failed delete retries and never moves on to create, and a retried create never deletes twice. An asset already gone from Ark counts as evicted. If a run gives up, the asset is left on Ark with no ledger row, and the hourly reconcile from #1519 deletes it.

**Ingest steps.** Per still: `-url` (fal key + fetchable URL, default retries) → `-claim` (the only minutes-long retry, and only for `pending`) → `-evict` → `-slot` → `-wait` → `-create`. Claim step name and result shape are unchanged so in-flight runs replay across the deploy.

**Parent timeouts.** Motion-batch and update-stale-shots await a motion child for 90 minutes (was 45 / the 30-minute default). Analyze-script awaits motion-batch for 120 minutes.

## Migrations
1. `byteplus_asset_leases`: new table.
2. `backfill_byteplus_asset_leases`: hand-written (data only). It copies each unexpired `lease_expires_at` to an owner `legacy:<slot id>`, so jobs polling across the deploy keep their stills.
3. `byteplus_asset_reservations` and follow-ups: `asset_id` becomes nullable by add `ark_asset_id`, copy, drop `asset_id` / `lease_expires_at`, then rename back — native ALTERs only (no table rebuild). The DROP COLUMNs are why `LEFTHOOK_EXCLUDE=migration-safety` was used. No foreign keys point at the table, so the #612 cascade cannot fire.

## Tests
- `byteplus-asset-pool.test.ts` (in-memory libsql through the real ledger SQL): last-slot CAS, same-still create-once, pending occupancy + `getAdmission`, eviction races, shared-still release, takeover/stale finalize, expired-reservation eviction, lease renewal, `pending` retryable vs `exhausted` `NonRetryableError`, `releaseOwner` of an unfinalized row.
- Motion: releases by owner on a fal success and in `onFailure`; ingest owner matches release owner. Studio: same. Steps: evict before the governor wait; a failed evict never reaches create.
- Not verified against the live BytePlus API.

https://claude.ai/code/session_01X434aTP5oU6u4cs56AfDSE


tombeckenham:
Closing to recreate the preview D1. Migrate failed with `duplicate column name: reserved_by` because this database already ran the first commit's table rebuild. Cleanup deletes `openstory-pr-1620`; I'll reopen once that's done.

tombeckenham:
Preview D1 `openstory-pr-1620` deleted (cleanup 200). Reopening so deploy-preview creates a fresh database and applies the native ALTER migrations from scratch.

